PR #1, opened 2016-09-17
Optimize for "small" maps (capacity fits in 32 bits)

bluss:
I think the case hash maps being smaller than 2<sup>32</sup> elements is pretty common.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,5 +19,9 @@ quickcheck = "0.3.1"
 fnv = "1.0"
 lazy_static = "0.2"
 
+[features]
+# for testing only, of course
+test_low_transition_point = []
+
 [profile.bench]
 debug = true
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,9 @@ use std::borrow::Borrow;
 
 use std::cmp::max;
 use std::fmt;
+use std::mem;
 use std::mem::{swap, replace};
+use std::marker::PhantomData;
 
 use util::{second, ptr_eq, enumerate};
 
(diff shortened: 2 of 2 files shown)


arthurprs:
I experimented with something similar yesterday so rebased and upload it to https://github.com/arthurprs/ordermap/commit/a74da54c75e6e98e9cb408ffeb6ce210d2476b25 for you to take a look.

In my case I assumed 1 trillion 99 billion (2^40-1) was a safe max size for any hash map and I only used the upper 24bits for the hash, making it very simple code wise but still able to save some lookups to the entries array.

```
➜  ordermap git:(master) ✗ cargo benchcmp old.txt new.txt 
 name                                   old.txt ns/iter  new.txt ns/iter  diff ns/iter   diff % 
 grow_fnv_orderedmap_10_000             260,540          273,855                13,315    5.11% 
 grow_orderedmap_10_000                 527,617          544,769                17,152    3.25% 
 insert_orderedmap_100_000              3,503,124        3,557,204              54,080    1.54% 
 insert_orderedmap_10_000               292,829          293,572                   743    0.25% 
 insert_orderedmap_150                  3,339            3,374                      35    1.05% 
 insert_orderedmap_int_bigvalue_10_000  570,576          569,547                -1,029   -0.18% 
 insert_orderedmap_str_10_000           349,569          349,155                  -414   -0.12% 
 insert_orderedmap_string_10_000        1,023,540        1,011,566             -11,974   -1.17% 
 iterate_orderedmap_10_000              4,616            4,472                    -144   -3.12% 
 lookup_orderedmap_100_000              1,611,977        1,555,304             -56,673   -3.52% 
 lookup_orderedmap_10_000               309,729          309,801                    72    0.02% 
 lookup_orderedmap_10_000_exist         183,007          155,581               -27,426  -14.99% 
 lookup_orderedmap_10_000_noexist       169,193          177,136                 7,943    4.69% 
 new_orderedmap                         3                3                           0    0.00% 
 with_capacity_10e5_orderedmap          4,167            4,039                    -128   -3.07%
```

bluss:
@arthurprs Very cool! I think the 32/32 split is better in some ways, since it allows using the truncated hash for the probe_distance calculation, so it's nicer to cache that way.

However, the branch on dispatching to the 32 or 64 bit size class shows up a lot in my code.

I agree with you that the larger the key, the more there's to win for ordermap.

bluss:
Creating a big map in a benchmark seems to make a mess of the timings. lazy_static seems to improve it a lot.

bluss:
I pushed the latest version of my branch. I'll have to compare benches with you, and see which approach scales better for larger maps

arthurprs:
Oh yeah, it still looking up the full hash for the probe_distance.
I'll check if I can make it smarter. As it is I'm sure it's no match for your optimization.

arthurprs:
Did you try to move the branch outside the probe_loop? It'll duplicate some code, but maybe the gains are worth it.

bluss:
I think neither approach can match the lookup for HashMap with u32 keys if we make the maps large enough unfortunately. 

My benchmark setup is not really that good (laptop), but it seemed to look like this when modifying the 100k lookup benches to use 1M keys instead:

```
lookup hashmap

test lookup_hashmap_1_000_000_multi          ... bench:     268,791 ns/iter (+/- 12,094)
test lookup_hashmap_1_000_000_single         ... bench:          91 ns/iter (+/- 4)

arthurps 40/24 split

test lookup_ordermap_1_000_000_multi         ... bench:     333,995 ns/iter (+/- 4,885)
test lookup_ordermap_1_000_000_single        ... bench:         110 ns/iter (+/- 11)

bluss 32 / 32 split

test lookup_ordermap_1_000_000_multi         ... bench:     316,870 ns/iter (+/- 5,147)
test lookup_ordermap_1_000_000_single        ... bench:          94 ns/iter (+/- 5)
```

It's as usual hard to find benchmarks that really answer the questions we have

bluss:
And, fill level policy determines a lot. It's quite sparse in ordermap in the current state (75%).

bluss:
I think the short hash branch is outside the loop. At least for .find_using_impl which is the method used by .get() (i.e. lookup). The size class case information is carried by the type parameter Sz there.

bluss:
Ok, this does seem like an improvement all around.

PR #2, opened 2016-10-11
Improve implementation of iterators

weiznich:
* match the implementation of the underlying iterators

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1039,6 +1039,18 @@ impl<'a, K, V> Iterator for Keys<'a, K, V> {
     fn size_hint(&self) -> (usize, Option<usize>) {
         self.iter.size_hint()
     }
+
+    fn count(self) -> usize {
+        self.iter.len()
+    }
+
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        self.iter.nth(n).map(|ent| &ent.key)
+    }
+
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1032,79 +1032,156 @@ pub struct Keys<'a, K: 'a, V: 'a> {
 impl<'a, K, V> Iterator for Keys<'a, K, V> {
     type Item = &'a K;
 
+    #[inline]
```
bluss: I would prefer to not specify this, it's the compiler's job. The item is already inlinable, so we don't need to do more. Same for all the other methods in the PR

src/lib.rs
```diff
@@ -1032,79 +1032,156 @@ pub struct Keys<'a, K: 'a, V: 'a> {
...
 }
 
 impl<'a, K, V> DoubleEndedIterator for Keys<'a, K, V> {
+    #[inline]
     fn next_back(&mut self) -> Option<&'a K> {
         self.iter.next_back().map(|ent| &ent.key)
     }
 }
 
+impl<'a, K, V> ExactSizeIterator for Keys<'a, K, V> {}
```
bluss: ESI is intentionally left out, because it's not really compatible with tombstones if that's the approach that will be adopted later.

weiznich:
Fixed

bluss:
Thanks a lot!

PR #6, opened 2016-11-02
Implement .retain()

bluss:


bluss:
It kind of validates the design if it's this easy to implement retain

PR #10, opened 2016-11-14
Abstract key lookup trait, so that it can be extended

bluss:


Techcable:
Could you please pull this? I can't implement Borrow for my type due to rust-lang/rust#39125

bluss:
I'm merging stuff into master preparing for version 0.3.0!

PR #12, opened 2016-11-15
Added Tombstones

Popog:
.retain() now returns an iterator with the removed items
closes bluss/ordermap#11
Also cleaned up some match statements.


bluss:
Thanks, that's interesting. It seems like a direct translation of the retain method into an iterator.

You're kind of putting me on the spot here, I have to review the design of the retain iterator and if I do it with my usual critical eye, there are some imperfections here:

- OrderMap gets an iterator that does not use the regular order of the elements
- Run on Drop is a questionable design pattern in itself. (Open question: Why should we use this pattern if we are making a new API?)
  - `forget` problematics show how nonintuitive it can be to tie behaviour to a temporary value
  - the operation is lazy as an iterator but eager on drop.
- I believe this crate can implement an in-order "retain" iterator with the same algorithmic complexity

Popog:
Well, that was why I said it seemed simple 😝 .

The only alternative I can think of to Run on Drop is how `thread::scoped` solved its issues. I'm certainly open to:
```rust
pub fn retain<F, G, R>(&mut self, keep: F, retain: G) -> R
    where F: FnMut(&mut K, &mut V) -> bool,
    G: FnOnce(Retain<K, V, S, F>) -> R
```

Which would complete the operation before returning.  I suspect `forget` would prevent, or at least significantly slow in-order traversal. So this would avoid that, and still allow users to do simple loops or `.collect()` it for storage.

bluss:
There's also the alternative of designing the operation to be entirely lazy, i.e. don't have any implicit action on drop, don't remove more elements than the iterator is driven into doing.

Popog:
That's somewhat unintuitive for users who don't want to iterate (i.e. people who would like to use the current function):
```rust
m.retain(|k, _| k < 5); // does nothing
m.retain(|k, _| k < 5).count(); // actually works
```

bluss:
Some ideas, just in that case:

1. Rename the method so that expectations change (can't keep the "retain" name since the expectations from existing APIs is wrong).
2. use `#[must_use]` so that it makes a warning when it does nothing (just like iterator adaptors like .map())
3. add an explicit method for draining the iterator / running it til the end.

Popog:
I think I'm now thoroughly convinced you cannot have all of these:
* O(n)
* A forgettable return value (no Drop or scoped style impl).
* No change to iteration order.

There maybe some exceedingly clever way of accomplishing this, but if so, I'm not seeing it.

bluss:
hehe, don't let me torture you for ordermap's purposes!

Popog:
But if you had to pick two...

Popog:
> 1. Rename the method so that expectations change (can't keep the "retain" name since the expectations from existing APIs is wrong).

Name idea `filter_cursor`

Popog:
Well, this no longer adds a `Retain` iterator, but it does add tombstones, which will make implementing `filter_cursor` trivially easy.

bluss:
Hah, crazy but interesting. Implementing tombstones is a big step. I'll have to look at this carefully. What are the benchmarks like?

Popog:
I've only changed `Entry` and `retain` to use tombstones, `swap_remove` and any benches which test that are unaffected. Overall, very little change. Tested on an i7 4770k:

name | original ns/iter | tombstone ns/iter
--- | --- | ---
entry_hashmap_150 | 3,209 (± 408) | 3,097 (± 60)
entry_orderedmap_150 | 2,846 (± 56) | 3,693 (± 48)
grow_fnv_hashmap_100_000 | 3,822,576 (± 134,932) | 3,862,308 (± 377,264)
grow_fnv_ordermap_100_000 | 4,027,347 (± 264,053) | 5,034,100 (± 735,180)
hashmap_merge_shuffle | 1,035,783 (± 49,163) | 1,064,027 (± 28,407)
hashmap_merge_simple | 34,034,802 (± 445,112) | 33,043,395 (± 355,684)
insert_hashmap_100_000 | 4,467,652 (± 1,126,408) | 4,277,086 (± 424,272)
insert_hashmap_10_000 | 267,512 (± 10,203) | 264,952 (± 7,215)
insert_hashmap_150 | 3,330 (± 63) | 3,220 (± 37)
insert_hashmap_int_bigvalue_10_000 | 644,812 (± 58,041) | 646,611 (± 13,821)
insert_hashmap_str_10_000 | 285,223 (± 5,886) | 284,730 (± 6,332)
insert_hashmap_string_10_000 | 1,229,139 (± 43,086) | 1,214,361 (± 22,182)
insert_orderedmap_100_000 | 3,656,966 (± 2,233,920) | 4,410,542 (± 811,701)
insert_orderedmap_10_000 | 251,699 (± 52,062) | 292,366 (± 15,493)
insert_orderedmap_150 | 3,269 (± 87) | 3,831 (± 19)
insert_orderedmap_int_bigvalue_10_000 | 458,839 (± 16,179) | 494,504 (± 13,260)
insert_orderedmap_str_10_000 | 265,618 (± 5,100) | 288,197 (± 2,769)
insert_orderedmap_string_10_000 | 1,083,413 (± 8,313) | 1,112,794 (± 14,495)
iter_black_box_hashmap_10_000 | 41,517 (± 572) | 44,543 (± 773)
iter_black_box_orderedmap_10_000 | 2,965 (± 24) | 5,509 (± 45)
iter_sum_hashmap_10_000 | 46,132 (± 1,379) | 44,069 (± 192)
iter_sum_orderedmap_10_000 | 2,594 (± 68) | 5,465 (± 101)
lookup_hashmap_100_000_inorder_multi | 164,799 (± 1,983) | 164,095 (± 1,107)
lookup_hashmap_100_000_multi | 170,325 (± 2,209) | 171,914 (± 5,477)
lookup_hashmap_100_000_single | 36 (± 8) | 35 (± 1)
lookup_hashmap_10_000_exist | 130,540 (± 8,240) | 127,333 (± 4,324)
lookup_hashmap_10_000_noexist | 135,121 (± 1,772) | 130,829 (± 1,057)
lookup_orderedmap_10_000_exist | 135,564 (± 2,158) | 145,417 (± 1,710)
lookup_orderedmap_10_000_noexist | 130,561 (± 2,740) | 136,118 (± 3,032)
lookup_ordermap_100_000_inorder_multi | 110,872 (± 1,311) | 117,463 (± 2,315)
lookup_ordermap_100_000_multi | 158,997 (± 3,455) | 159,648 (± 1,922)
lookup_ordermap_100_000_single | 33 (± 33) | 33 (± 10)
new_hashmap | 7 (± 0) | 7 (± 0)
new_orderedmap | 11 (± 0) | 14 (± 0)
ordermap_merge_shuffle | 646,260 (± 16,298) | 977,019 (± 5,665)
ordermap_merge_simple | 430,928 (± 36,904) | 786,512 (± 7,431)
pop_ordermap_100_000 | 2,989,304 (± 151,554) | 2,588,539 (± 49,431)
remove_ordermap_100_000 | 6,585,708 (± 447,388) | 5,901,486 (± 1,301,382)
retain_few_ordermap_100_000 | 3,533,252 (± 132,025) | 2,266,023 (± 67,881)
retain_half_ordermap_100_000 | 3,127,368 (± 608,248) | 2,035,521 (± 39,373)
retain_many_ordermap_100_000 | 1,127,635 (± 115,684) | 1,299,009 (± 22,572)
with_capacity_10e5_hashmap | 5,150 (± 148) | 5,115 (± 19)
with_capacity_10e5_orderedmap | 5,454 (± 130) | 5,485 (± 47)
insert_hashmap_string_10_000 | 1,257,354 (± 60,714) | 1,229,721 (± 7,204)
insert_hashmap_string_oneshot_10_000 | 1,173,116 (± 67,376) | 1,149,863 (± 5,214)
insert_orderedmap_string_10_000 | 1,093,339 (± 23,124) | 1,092,168 (± 13,019)
lookup_hashmap_10_000_exist_string | 190,106 (± 2,865) | 190,600 (± 2,763)
lookup_hashmap_10_000_exist_string_oneshot | 171,826 (± 54,066) | 170,334 (± 2,971)
lookup_ordermap_10_000_exist_string | 217,790 (± 4,591) | 224,351 (± 1,487)
lookup_ordermap_10_000_exist_string_oneshot | 197,931 (± 7,179) | 192,204 (± 2,247)

I'm going to investigate using the MSB of the hash in place of `Option (requires `unsafe` [either directly or as a dependency]) as I have a hunch that will probably provide some perf benefits.

Popog:
Found some bugs and fixed them. Also added some unsafe options under test feature flags.  They don't really seem to provide any perf benefit, so I'll probably end up scrapping them.

The CI break seems to be from an SSL error for crates.io trying to pull the tagged union packaged I added as a dependency. Not sure what's up with that but with how things are looking I'll probably end up removing the dependency.

Popog:
So, found a couple more bugs, but while I'm already touching so much of the code, how would you feel about the following changes:
* Remove `self.mask` in favor of getting the mask via `indices.len() - 1`
* Removing `probe_loop!` in favor of `loop` and `probe = (probe + 1) & mask`

My justification for the former is, it seems like a bad idea to have duplicate data that could get out of sync. The mask is already implicitly stored in `indices` and is cheap to recover.

My justification for the later is it reads better than the macro (especially when debugging as line numbers get collapsed to the first line of the macro) and seems to have very little perf cost.

bluss:
I'm sorry I haven't sat down and looked at this yet. I would hold off on both of those changes.

Popog:
No problem, I've had to do quite a bit of cleanup before this PR would actually be usable. I've removed all my unsafe testing (it ultimately worked out to a maybe a 5% gain) and switched everything over to tombstones which can benefit from them.

bluss:
I'm sorry that I haven't been able to review this. I think tombstones are going to be useful, and maybe we should bench it against linked hash map instead for a more apples to apples comparison. Unsure if we should fork the crate for this, I almost think so if we can't bring the overhead down. Improving retain just to have lookup and insert regress really is throwing the baby out with the bath water for a general purpose hash table.

bluss:
Ok, I'm clearly not fit to help this PR as a maintainer. I would recommend forking the crate and giving the map a different name, because with tombstones you get a different hash table with other nice properties, while removing some of the nice properties of this OrderMap. (Performance and compactly indexed set).

bluss:
@Popog Hey! What do you think about publishing ordermap with full insertion order (using tombstones) as a separate crate?

Popog:
Got a good idea for a name? (I'm the worst at naming things)

bluss:
A. What if current ordermap is renamed to donate its name?
Current ordermap is characterized by (1) consistent hash-independent order (2) compactly indexable map, so what's a good name for it?? Now we have the same problem... maybe "CompactMap"?

B. ordermap stays and the new type with matching crate is "InsertionOrderMap"?

bluss:
compactmap is taken: https://crates.io/crates/compactmap

bluss:
I think this PR would be simpler if it didn't use index tombstones, only entry tombstones. We don't need to keep any index order

Popog:
Yeah, the index tombstones was just an attempt to amoratize other operations. I should have time this weekend to go back a couple commits to just entry tombstones and clean that up.

What about instead of a new crate, just adding it as a another type to this crate?

bluss:
I'd prefer to split the crates if we are not at the point where implementation can be shared. I think all the core probe loops etc will look different, so there will not be so much actual overlap?

There's also the reason that if it's in this crate I'd feel responsible for it, and I don't have time to be responsible for another one unfortunately.

cuviper:
I believe this PR is obsolete since renaming to `indexmap`.

I think it's been long enough that we could donate / give up the `ordermap` name now, if that's still desired. Let us know if you want it.

PR #13, opened 2017-02-13
Implement reserve and shrink_to_fit

Popog:
Repurposed double_capacity to change_capacity.

Also added a .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+target
+Cargo.lock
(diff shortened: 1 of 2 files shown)


Amanieu:
Any update on this? I would like to use `shrink_to_fit` after I am done building a map to reduce its memory usage. The map is very large (>1M entries) is used for the entire lifetime of the program.

cuviper:
#131 implemented the same in the switch to hashbrown.

PR #15, opened 2017-02-20
Add ordered_remove with `O(n)` perf.

cristicbz:
I might be understanding the implementation wrong, but I think that order-preserving remove can be implemented by shifting both the `indices` and entry `Vec`-s, then decrementing all indices after the removed one, then correcting the robin hood property. But maybe I'm wrong, please review carefully @bluss :)

Because this operation is `O(n)` compared to the std `O(1)`, I named it `ordered_remove` so that you're forced to make a conscious decision to preserve order at the expense of performance.

If this implementation makes sense, I'm happy to add a couple more commits to update the README and add this operation to `Entry` too.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -165,6 +165,21 @@ impl Pos {
             }
         }
     }
+
+    #[inline]
+    fn update_index<Sz, F>(&mut self, updater: F)
+        where Sz: Size,
+              F: FnOnce(usize) -> usize
+    {
+        if !self.is_none() {
+            if Sz::is_64_bit() {
+                self.index = updater(self.index as usize) as u64;
+            } else {
+                let (index, hash) = split_lo_hi(self.index);
+                self.index = updater(index as usize) as u64 | ((hash as u64) << 32);
+            }
+        }
+    }
 }
 
 #[inline]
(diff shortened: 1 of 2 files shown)


bluss:
This does look good to me, just not sure if we need it. I *think* it is ultimately needed, the OrderMap has a lot of Vec like functionality going already.

bluss:
Since some time has passed, you can disregard review comments here. But I'll put down my thoughts so that I don't forget them, for when I come back to this.

src/lib.rs
```diff
@@ -165,6 +165,21 @@ impl Pos {
...
+    fn update_index<Sz, F>(&mut self, updater: F)
+        where Sz: Size,
+              F: FnOnce(usize) -> usize
+    {
+        if !self.is_none() {
+            if Sz::is_64_bit() {
+                self.index = updater(self.index as usize) as u64;
+            } else {
+                let (index, hash) = split_lo_hi(self.index);
+                self.index = updater(index as usize) as u64 | ((hash as u64) << 32);
```
bluss: thought: does this truncate index correctly

src/lib.rs
```diff
@@ -1074,36 +1109,35 @@ impl<K, V, S> OrderMap<K, V, S> {
...
-                        self.indices[probe] = Pos::with_hash::<Sz>(found, entry.hash);
-                        break;
-                    }
+        for index in self.indices.iter_mut() {
+            index.update_index::<Sz, _>(|entry_index| {
+                if entry_index >= found {
+                    entry_index - 1
+                } else {
+                    entry_index
                 }
```
bluss: If found == 0; entry_index - 1 wraps around

bluss:
Ok, to not bother you further, I've copied this branch for later reuse to this repo and I'll close the PR. :-)

Issue #19, opened 2017-02-28
Full Consistency With HashMap


bluss:
### Missing methods (stable):

- [x] `with_hasher` #35
- [x] `hasher` #35 
- [x] `shrink_to_fit`
- [x] `is_empty` #22 
- [x] `drain` #27 
- [x] `values`, `values_mut` #17 

### New methods:

**As a rule, we don't implement unstable methods; they are implemented when they are stable in Rust beta**

- [x] `get_key_value` rust-lang/rust/issues/49347
- [x] `Entry::and_modify` (unstable)
- [x] `Entry::or_default` (unstable)
- [x] `try_reserve` rust-lang/rust/issues/48043
- [x] `remove_entry` rust-lang/rust/issues/46344

### Trait impls

- [x] `PartialEq + Eq` #23 
- [x] `Extend<(&K, &V)>` #34 
- [x] Ser/De #28 
- [x] ESI for iterators #26 
- [x] Debug for Entry
- [x] Clone for some iterators (Keys, Values, Iter, probably all of them)

### Type Conflicts:

- [x] `Entry<K, V>` vs `Entry<K, V, S>`  (not important) #59
- [x] `retain`'s `&K` vs `&mut K` #39 
- [ ] `drain`'s signature (not important)
- [x] `OccupiedEntry::insert`'s signature #68

xfix:
#28 implements Ser/De

bluss:
I can't find any discrepancies between HashSet and OrderSet. A second close look is of course welcome, but I think @cuviper already did this carefully when implementing OrderSet.

CAD97:
`indexmap::map::Keys` is not `Clone` where `std::collections::hash_map::Keys` is. (Just noticed this moving some code that needed stable ordering over.) Is this an innate limitation or just an oversight?

bluss:
Just an oversight, please check all iterators you can and PRs welcome.

cuviper:
I think we are as close as we can get at this point. New additions to `std` can continue being considered individually.

PR #23, opened 2017-03-23
Implement PartialEq and Eq

overvenus:
Hello again!

This PR implements traits `PartialEq` and `Eq`. :grinning: 

Cc #19

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1463,12 +1463,13 @@ impl<K, V, S> Default for OrderMap<K, V, S>
     }
 }
 
-impl<K, V, S> PartialEq for OrderMap<K, V, S>
-    where K: Eq + Hash,
-          V: PartialEq,
-          S: BuildHasher
+impl<K, V1, S1, V2, S2> PartialEq<OrderMap<K, V2, S2>> for OrderMap<K, V1, S1>
+    where K: Hash + Eq,
+          V1: PartialEq<V2>,
+          S1: BuildHasher,
+          S2: BuildHasher
 {
-    fn eq(&self, other: &OrderMap<K, V, S>) -> bool {
+    fn eq(&self, other: &OrderMap<K, V2, S2>) -> bool {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1463,6 +1463,27 @@ impl<K, V, S> Default for OrderMap<K, V, S>
     }
 }
 
+impl<K, V, S> PartialEq for OrderMap<K, V, S>
+    where K: Eq + Hash,
```
bluss: I'd like to do this more generally than libstd.

We sometimes have room to fix their mistakes or oversights, right? And we should do it if it remains a drop in replacement.

What I first thought of here is that PartialEq should not require that both maps have the same hasher. But in fact, we can generalize it so that neither K, V, S need to be the same as long as they implement `V1: PartialEq<V2>` and so on. Does that make sense to you?

For `K` it's a bit trickier, we need to be able to look up using one key in the other map. I'm not sure what works there without trying it out in the code.

overvenus: That is a good idea!

For `K`, my first thought is `std::convert` traits and `Borrow`, but later on I found none of them are suitable here. `PartialEq` requires **symmetric**, a == b implies b == a. I'm not sure what is best for `K`, maybe they should be identical. :thinking:

bluss: Hm we can wait a bit with K and merge the Equivalent PR. It's a generalization of the Borrow mechanism https://github.com/bluss/ordermap/pull/10

bluss:
I'll have to experiment with the K issue

bluss:
Hm we can merge this since the Equivalent trait is waiting on the next version anyway, so we can do a small type inference break then.

bluss:
PartialEq should be exercised in quickcheck tests. Ideally you'd just add a new one next to the ones that test equivalence with HashMap. This test can be added by you or I, depending on who gets there first.

PR #27, opened 2017-04-20
Adds a OrderMap::drain() method

stevej:
Problem:
Porting some code that used HashMap's drain to OrderMap revealed a small
API gap.

Solution:
Added a drain method, mostly mirroring what std::collections::HashMap
offers except returning a standard Vec instead of a Drain.

drain() provides an efficient way to remove all values without shrinking
the capacity of the original OrderMap.

Validation:
Added unit tests and ported an existing HashMap::drain() user to
OrderMap::drain

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -400,6 +400,18 @@ impl<K, V, S> OrderMap<K, V, S>
     pub fn capacity(&self) -> usize {
         usable_capacity(self.raw_capacity())
     }
+
+    /// Clears the `OrderMap`, returning all key-value pairs as a `Drain`.
+    /// Keeps the allocated memory for reuse.
+    pub fn drain(&mut self) -> Drain<K, V> {
+        for i in &mut self.indices {
+            *i = Pos::none();
+        }
+
+        Drain {
+            inner: self.entries.drain(..),
+        }
+    }
 }
 
 /// Trait for the "size class". Either u32 or u64 depending on the index
@@ -1496,6 +1508,17 @@ impl<K, V, S> Eq for OrderMap<K, V, S>
 {
 }
 
+pub struct Drain<'a, K, V> where K: 'a, V: 'a {
+    inner: ::std::vec::Drain<'a, Bucket<K, V>>
+}
+
+impl<'a, K, V> Iterator for Drain<'a, K, V> {
+    type Item = (K, V);
+    fn next(&mut self) -> Option<Self::Item> {
+        self.inner.next().map(|bucket| (bucket.key, bucket.value))
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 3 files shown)


carllerche:
I'd probably try to avoid requiring the allocations as part of `drain`.

I'd suggest adding a new `Drain` type. Something like:

```
struct Drain<'a, K, V> {
    inner: std::vec::Drain<'a, Bucket<K, V>>,
}
```

Then, impl `Iterator` on drain to map out `(K, V)` in `Iterator::next`.

I'd probably also try to avoid the `vec!` allocation when resetting `indices`. Iterating the list and setting each entry to `Pos::none()` should be sufficient.

stevej:
Thank you, Carl! I rewrote the code to address those concerns. Please let me know if I made a mistake, I haven't implemented that many Iterators in Rust.

src/lib.rs
```diff
@@ -400,6 +400,18 @@ impl<K, V, S> OrderMap<K, V, S>
     pub fn capacity(&self) -> usize {
         usable_capacity(self.raw_capacity())
     }
+
+    /// Clears the `OrderMap`, returning all key-value pairs as a `Drain`.
```
bluss: Iterators don't really have a type-identity, they are peculiar that way. The struct is just a vehicle for the implementation, so we don't call it a `Drain` but maybe a drain iterator.

I think that the method should move down the method list to be placed at a location that degrades it in importance and where users more expect it... after the other iterator methods.

src/lib.rs
```diff
@@ -1496,6 +1508,17 @@ impl<K, V, S> Eq for OrderMap<K, V, S>
...
+pub struct Drain<'a, K, V> where K: 'a, V: 'a {
+    inner: ::std::vec::Drain<'a, Bucket<K, V>>
+}
+
+impl<'a, K, V> Iterator for Drain<'a, K, V> {
+    type Item = (K, V);
+    fn next(&mut self) -> Option<Self::Item> {
+        self.inner.next().map(|bucket| (bucket.key, bucket.value))
+    }
+}
```
bluss: Please add `size_hint`, it's the next important iterator method. If the other OrderMap iterators are exact size iterators, this one should be too. I'm not too exited for any other iterator method implementations, we don't necessarily need those.

src/lib.rs
```diff
@@ -1690,4 +1713,39 @@ mod tests {
         assert_ne!(map_a, map_c);
         assert_ne!(map_c, map_a);
     }
+
+
+    #[test]
+    fn drain_basic() {
+        let mut map_a = OrderMap::new();
```
bluss: (Not very actionable comment)

Quickcheck tests are always better than example tests. In this case we reuse a solid std iterator and we don't really *need* any particular test.

carllerche:
@bluss the main problem w/ taking an argument right now is that `RangeArgument` isn't stable... maybe name the fn `drain_all` for now and reserve `drain` for when the trait does hit stable? Then deprecate `drain_all`.

bluss:
unstable RangeArgument is not a problem. Simply provide an equivalent trait. (Arrayvec does this).

bluss:
Even less of a problem is doing the simple stopgap solution which is to take a single `..` as an argument in the initial version.

src/lib.rs
stevej: I've replaced the example unit tests with a drain quickcheck test. This is my first time using Rust's quickcheck, please let me know if you think I made some newbie error.

src/lib.rs
stevej: Added by delegating to `inner.size_hint()`

src/lib.rs
stevej: method moved.

stevej:
Let me know if you'd rather I squashed this into a single logical commit.

bluss:
Thanks!

bluss:
This checks off an item on HashMap consistency list but adds another (That's my meddling).

PR #28, opened 2017-05-18
Implement optional Serde 1.0 support

xfix:
Note that this is not ready for merging, as it doesn't run Travis tests. I'm not sure what should be testing strategy?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ description = "A hash table with consistent order and fast iteration."
 bench = false
 
 [dependencies]
+serde = { version = "1.0", optional = true }
 
 [dev-dependencies]
 itertools = "0.5.1"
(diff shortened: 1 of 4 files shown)


bluss:
One can use a private crate in the repo with extra tests if one wants to split out those tests and their dependencies.

Thanks for working on this.

I kind of want to say that we need to support both kinds of serialization with serde (the kind that gets context from the stream [JSON] and the bon-self-describing kind [bincode]). Unfortunately I'm not 100% sure how to ensure compat with both.

xfix:
This I'm fairly sure should support both JSON and bincode. All the serialization method does is telling that this is a hash map, and its size (for formats that require size in advance, like bincode). The deserialization method is really simple, it just deserializes a map with size that may or may not be specified by a format (JSON doesn't specify map size, bincode does).

The test function tests if this can be successfully serialized and deserialized.

LegNeato:
Any chance on getting this merged soon?

bluss:
Sure. I'm sorry that it fell of my to do list. Friendly pings are good.

bluss:
I'm fixing up .travis.yml

bluss:
In 0.2.11

LegNeato:
Wow, thanks for the quick turnaround on the merging and release!

PR #32, opened 2017-08-15
Add an example to the crate level docs

dns2utf8:
It is just a simple example. I am open to suggestions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,38 @@
+//! Experimental hash table implementation in just Rust (stable, no unsafe
+//! code).
+//!
+//! This is inpired by Python 3.6's new dict implementation (which remembers
+//! the insertion order and is fast to iterate, and is compact in memory).
+//!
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1,3 +1,38 @@
+//! Experimental hash table implementation in just Rust (stable, no unsafe
+//! code).
+//!
+//! This is inpired by Python 3.6's new dict implementation (which remembers
+//! the insertion order and is fast to iterate, and is compact in memory).
+//!
+//! This implementation corresponds to their "compact hash map" idea as it is
+//! now, and has consistent order.
```
bluss: Thanks for the PR. I understand that the crate is confusingly documented as it is, but rustdoc is where it should be correct. The ordering explanation on the struct OrderMap itself is better and more correct. Resting on the reference to Python makes it easy to misunderstand what it actually does.

bluss:
closing -- feel free to talk to me on irc or somewhere and we can add an example

PR #34, opened 2017-08-29
Implement Extend<(&K, &V)> for Copy types

xfix:
This matches `hash_extend_copy` feature stabilized in Rust 1.4.0 introduced by https://github.com/rust-lang/rust/pull/28094. Note that this may affect type inference.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1517,6 +1517,16 @@ impl<K, V, S> Extend<(K, V)> for OrderMap<K, V, S>
     }
 }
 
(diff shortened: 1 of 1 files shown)


bluss:
We're starting breaking changes now. Not that this is so much to worry about, but why not. A few minor things together and we might as well.

bluss:
Thank you for the PR!

bluss:
Related to #19

bluss:
Finally in ordermap 0.3.0

bluss:
(As usual, say if the changelog credit should be something else than the github nickname)

PR #36, opened 2017-09-01
Link the README.md file in Cargo.toml

notriddle:
This way, crates.io and docs.rs will render it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ documentation = "https://docs.rs/ordermap/"
 repository = "https://github.com/bluss/ordermap"
 license = "Apache-2.0/MIT"
 description = "A hash table with consistent order and fast iteration."
+readme = "README.md"
 
 [lib]
 bench = false


bluss:
I'm sorry, there is no README.md file :smile:

notriddle:
Oh, oops.

...

I guess if I set `readme = "README.rst"`, crates.io will try to render the ReStructuredText as Markdown? @anthonynguyen

bluss:
Yes. It doesn't look entirely terrible but it's not so good.

bluss:
Oh we have an example from ndarray 0.10.4 before I removed the readme key. For it, the rst as markdown looks especially bad

https://crates.io/crates/ndarray/0.10.4

notriddle:
:-1: nope 😞

bluss:
Thanks anyway! I'm impressed by the diligence of the rust community, so many people were ready with their readme keys long before it had any use.

PR #39, opened 2017-09-24
Add trait MutableKeys

### LINKED ISSUE #7: Move mutable key access to a trait
We should put this feature a bit on the side, in a trait.

Something like:

```rust
impl Ordermap {
    fn get_pair_mut(&mut self, key: &Q) -> (&K, &mut V) { .. }
}

impl MutableKeys for OrderMap {
    fn good_method_name(&mut self, key: &Q) -> (&mut K, &mut V) {  .. }
}
```

bluss:
- Remove get_pair, and rename get_pair_index to get_full. Same with the
  mutable variant.

- Move mutable key methods to MutableKeys -- we want this feature to be
  out of the way, but opt-in. The trait has mutable key-versions of
  get_full and retain

Fixes #7

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,7 +19,7 @@ use std::fmt;
 use std::mem::{replace};
 use std::marker::PhantomData;
 
-use util::{second, third, ptrdistance, enumerate};
+use util::{third, ptrdistance, enumerate};
 pub use equivalent::Equivalent;
 pub use mutable_keys::MutableKeys;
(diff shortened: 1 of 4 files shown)


bluss:
cc @garro95 github code search found you as a user of mutable keys. As a head's up, ordermap is preparing some small remodeling to put mutable key methods on a trait, so that users opt in to a "nonstandard" interface.

PR #40, opened 2017-09-26
Added index() method for OccupiedEntry

### LINKED ISSUE #31: Feature request: Entry::get_index
In the Entry API, OccupiedEntry should expose a function to get the index of the entry.

garro95:
This fix #31. Easy :-)


bluss:
Thanks! I think it would be good to follow up with index methods on VacantEntry and then Entry itself too, I think. This rounds off and completes our API to make it more consistent.

One can spy the right value to use for VacantEntry from the line `let index = self.map.entries.len();` that might be replaced by using the new method in fact.

PR #41, opened 2017-09-29
index() method for Entry and VacantEntry plus test coverage for Entry API

garro95:
I implemented the `index()` method for `VacantEntry` that just returns `self.map.len()` and for `Entry`.
I also added a test for some of the functions provided by the Entry API


bluss:
Looks great, thanks a lot!

PR #42, opened 2017-09-30
Fix serde deserialization for custom hashers.

Techcable:
Although serialization was implemented, deserialization wasn't.
I've added a test for fnv hashers to prevent any regressions ;)

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -27,13 +27,14 @@ impl<K, V, S> Serialize for OrderMap<K, V, S>
     }
 }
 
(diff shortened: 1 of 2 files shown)


bluss:
restarted the travis job that was hanging unstarted

bluss:
Completed and passed. Thank a lot for getting back to this -- at some point I had to double check that we rembembered S, but apparently I only looked at Serialization too.

This is worthy of a patch release to ordermap branch 0.2, but I can do that, just cherry pick your changes after they are merged to master.

PR #45, opened 2017-11-20
Implement Rayon traits (optional)

cuviper:
- `IntoParallelIterator` by value, ref, and ref mut
  - Implements both `ParallelIterator` and `IndexedParallelIterator`
- `FromParallelIterator`
- `ParallelExtend` by value and ref copy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ matrix:
     - rust: stable
       env:
        - FEATURES='serde-1'
+    - rust: stable
+      env:
+       - FEATURES='rayon'
     - rust: beta
     - rust: nightly
     - rust: nightly
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,4 +54,4 @@ debug = true
 no-dev-version = true
 
 [package.metadata.docs.rs]
-features = ["serde-1"]
+features = ["serde-1", "rayon"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ bench = false
 
 [dependencies]
 serde = { version = "1.0", optional = true }
+rayon = { version = "1.0", optional = true }
 
 [dev-dependencies]
 itertools = "0.8"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,6 +23,8 @@
 
 #[macro_use]
 mod macros;
+#[cfg(feature = "rayon")]
+mod rayon;
 #[cfg(feature = "serde-1")]
 mod serde;
 mod util;
@@ -78,3 +80,9 @@ impl<K, V> Bucket<K, V> {
     fn muts(&mut self) -> (&mut K, &mut V) { (&mut self.key, &mut self.value) }
 }
 
+trait Entries {
+    type Entry;
+    fn into_entries(self) -> Vec<Self::Entry>;
+    fn as_entries(&self) -> &[Self::Entry];
+    fn as_entries_mut(&mut self) -> &mut [Self::Entry];
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,8 +23,6 @@
 
 #[macro_use]
 mod macros;
-#[cfg(feature = "rayon")]
-mod rayon;
 #[cfg(feature = "serde-1")]
 mod serde;
 mod util;
@@ -33,6 +31,11 @@ mod mutable_keys;
 
 pub mod set;
 pub mod map;
+
+// Placed after `map` and `set` so new `rayon` methods on the types
+// are documented after the "normal" methods.
+#[cfg(feature = "rayon")]
+mod rayon;
 
 pub use equivalent::Equivalent;
 pub use map::IndexMap;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -85,4 +85,6 @@ trait Entries {
     fn into_entries(self) -> Vec<Self::Entry>;
     fn as_entries(&self) -> &[Self::Entry];
     fn as_entries_mut(&mut self) -> &mut [Self::Entry];
+    fn with_entries<F>(&mut self, f: F)
+        where F: FnOnce(&mut [Self::Entry]);
 }
--- a/src/map.rs
+++ b/src/map.rs
@@ -4,7 +4,8 @@
 pub use mutable_keys::MutableKeys;
 
 #[cfg(feature = "rayon")]
-pub use rayon::map::{IntoParIter, ParIter, ParIterMut};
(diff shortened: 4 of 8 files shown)


src/rayon.rs
```diff
@@ -0,0 +1,209 @@
...
+// This form of intermediate collection is also how Rayon collects `HashMap`.
+// Note that the order will also be preserved!
+fn collect<I: IntoParallelIterator>(iter: I) -> LinkedList<Vec<I::Item>> {
+    iter.into_par_iter()
+        .fold(Vec::new, |mut vec, elem| {
+            vec.push(elem);
+            vec
+        })
+        .collect()
+}
```
bluss: Loll wut!! LinkedList is used in a real-life PR?!! 🥇 🐫 🎈 :shipit: 

(Ahem I'll be back to my regular @bluss in a bit.)

src/rayon.rs
```diff
@@ -0,0 +1,209 @@
...
+            where C: UnindexedConsumer<Self::Item>
+        {
+            self.entries.into_par_iter()
+                .map($map_elt)
+                .drive_unindexed(consumer)
+        }
+
+        // NB: This allows indexed collection, e.g. directly into a `Vec`, but the
+        // underlying iterator must really be indexed.  We should remove this if we
+        // start having tombstones that must be filtered out.
```
bluss: Great comment.

src/rayon.rs
cuviper: You know, it's funny, for all the strategies we benchmarked for Rayon collect, I don't think we actually tried a `Vec<Vec<T>>` intermediate.  I just assumed that a list was best for this purpose because of its O(1) append.  I'm going to have to do some testing now...

cuviper: I've kept the `LinkedList`, but now collect it with map-reduce -- see rayon-rs/rayon#479.

bluss:
As awesome as this is, it does interfere with easy 1.0'ing of ordermap.

cuviper:
Would you want to wait for rayon-1.0?  Or would you hesitate even then?

Another idea is to add `as_slice()`, `as_mut_slice()`, and `into_vec()`, and put a public API on the `Bucket`, so people could themselves write something like:

```rust
omap.as_slice().into_par_iter()
    .map(|b| (b.key(), b.value())))
    .for_each(|(k, v)| ...);
```

That's not as ergonomic, but at least it would be *possible* to use Rayon with ordermap efficiently, and maybe enable other cool uses too!
(tombstones would make this uglier though.)

bluss:
With rayon-1.0 I would not hesitate, but you're right, one could be worried about the version entanglement even then.

For the record, we've decided that OrderMap will stay as it is with its current properties, indexing, exact size = no tombstones (and probably the same name too, if we don't find a better one). This data structure is useful as it is, so to implement a similar one with tombstones, it will be a new thing. We could end up with both.

bluss:
W.r.t order I'm very fond of a deterministic order (hash function independent).

I had a different thought -- would it be possible to implement OrderMap shared and mutable views. So that it would natively support splitting (by index) into two parts. Both the shared and mutable view would mean shared/mutable access to the elements only, not inserting or removing elements. I think that means that the mutable views can all have the same shared immutable reference to the index/hash vector.

In a mutable view, the implementation peeking into another view's part of the entries vector would be verboten.

cuviper:
I think views are a neat idea!  It wouldn't directly enable rayon, but it could let there be something like `ordermap-parallel`, like you did with `ndarray` and `ndarray-parallel`.

vitiral:
Just a thought: IMO feature flags should be able to be independent of the crate stability. I.e. you can advertise that "rayon is supported but is beta".

bluss:
Yes, maybe. I think the feature flag must have unstable in its name for it to work. It's different viewpoints among users in the role of docs vs formal API in defining the stable interface.

cuviper:
I've now rebased, updated to rayon 1.0, and added the basic set implementations.

bluss:
That's cool! I'm fine with merging this, but I haven't looked through it in fine detail. rayon 1.0 is quite the thing, and maybe we can dare to go 1.0 with it as dependency. (Congratulations!)

cuviper:
I was also thinking of `par_eq`, `par_is_disjoint`, `par_is_subset`, and `par_is_superset`, as these are all operations that have to (potentially) iterate the whole set.  This might be too much `par_` though. 😜 

The only thing left would be `drain`, which we haven't really addressed at all in rayon.  I'm thinking about adding a general `ParDrain` trait, but I'm not sure.

Anyway, I should at least write some documentation and tests here before we merge. 😅 

One specific API question for you is whether it's good to have these `Par`* types re-exported in the main `map` and `set` modules, next to their serial counterparts.  It might be better in separate modules, perhaps even public from the `indexmap::rayon::map`/`set` where they're currently implemented, and remove `Par` from their names. (like Rayon does for `std` iterator types)

bluss:
This PR is clearly amazing. I think this is a great feature set for a hashmap library and in safe Rust no less (thanks to the giants of std::vec::Vec and rayon).

I think that's a lot of functions. I'd probably drop the comparisons and `par_sorted_by` (`sorted_by` would probably not have existed if `sort_by` was realized first? It is a tad more efficient, than `sort_by` though..). I'm just wondering if we can get most of the value with less API.

I don't know about the module and name prefix question. Mirroring std fits well with rayon. Here I don't mind either way.

ConnorGray:
What's the status of this pull request? I'm a fan of IndexMap, but right now I'm in a situation where I'd really like to be able to use rayon to get a performance boost, and unfortunately can't.

ConnorGray:
@cuviper Wow the quick response is much appreciated :)

I don't have a much input -- I'm just hoping to take advantage of rayon's accessible parallelism (where not having to worry about the details is a selling point) with the determinism I get from using an IndexMap everywhere internally.

I imagine there might be others in the same boat, who like IndexMap for it's determinism, but would also like to operate on the values in parallel.

bluss:
I think this is good. Rayon is rock solid :wink: and doesn't feel like a risk. The code is simple.

Only review comment that is important to fix, is that the new methods (par_keys etc) need to be listed last in the docs, not first in the docs. I'd enable the rayon feature in docs.rs generation and make sure items are labelled in docs to require that feature (I can do that).

garro95:
What's the state of this? I would like to use this feature

cuviper:
I think this is ready for reconsideration.

cuviper:
OK, that conflict was just in the compat downgrades, which #97 removed.

Rebased without that, and CI passed -- r=bluss

bluss:
Great. I can do the release of 1.1.0

PR #46, opened 2017-11-21
Add OrderSet

### LINKED ISSUE #8: Feature Request: OrderSet?
Like `HashSet`, but with this crate's hash table under the hood

cuviper:
An `OrderSet<T>` wraps an `OrderMap<T, ()>`, and also adds set-like
functionality: union, intersection, etc.

The type and all of its supporting iterators are defined in the public
`set` module, and `OrderSet` is re-exported at the crate root.  There is
also a new `orderset!` macro for construction.

Fixes #8.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -59,10 +59,13 @@ macro_rules! ordermap {
 /// # }
 /// ```
 macro_rules! orderset {
+    (@single $($x:tt)*) => (());
+    (@count $($rest:expr),*) => (<[()]>::len(&[$(orderset!(@single $rest)),*]));
+
     ($($value:expr,)+) => { orderset!($($value),+) };
     ($($value:expr),*) => {
         {
-            let _cap = ordermap!(@count $($value),*);
+            let _cap = orderset!(@count $($value),*);
             let mut _set = $crate::OrderSet::with_capacity(_cap);
             $(
                 _set.insert($value);
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -36,3 +36,38 @@ macro_rules! ordermap {
         }
     };
 }
+
+#[macro_export]
+/// Create an `OrderSet` from a list of values
+///
+/// ## Example
+///
+/// ```
+/// #[macro_use] extern crate ordermap;
+/// # fn main() {
+///
+/// let set = orderset!{
+///     "a",
+///     "b",
+/// };
+/// assert!(set.contains("a"));
+/// assert!(set.contains("b"));
+/// assert!(!set.contains("c"));
+///
+/// // "a" is the first value
+/// assert_eq!(set.iter().next(), Some(&"a"));
+/// # }
+/// ```
+macro_rules! orderset {
+    ($($value:expr,)+) => { orderset!($($value),+) };
+    ($($value:expr),*) => {
+        {
+            let _cap = ordermap!(@count $($value),*);
+            let mut _set = $crate::OrderSet::with_capacity(_cap);
+            $(
+                _set.insert($value);
+            )*
+            _set
+        }
+    };
+}
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -74,3 +74,39 @@ macro_rules! orderset {
         }
     };
 }
+
+// generate all the Iterator methods by just forwarding to the underlying
+// self.iter and mapping its element.
+macro_rules! iterator_methods {
+    // $map_elt is the mapping function from the underlying iterator's element
+    // same mapping function for both options and iterators
+    ($map_elt:expr) => {
+        fn next(&mut self) -> Option<Self::Item> {
+            self.iter.next().map($map_elt)
+        }
+
+        fn size_hint(&self) -> (usize, Option<usize>) {
+            self.iter.size_hint()
+        }
+
+        fn count(self) -> usize {
+            self.iter.len()
+        }
+
+        fn nth(&mut self, n: usize) -> Option<Self::Item> {
+            self.iter.nth(n).map($map_elt)
+        }
+
+        fn last(mut self) -> Option<Self::Item> {
+            self.next_back()
+        }
+
+        fn collect<C>(self) -> C
+            where C: FromIterator<Self::Item>
+        {
+            // NB: forwarding this directly to standard iterators will
+            // allow it to leverage unstable traits like `TrustedLen`.
+            self.iter.map($map_elt).collect()
+        }
+    }
+}
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -43,7 +43,7 @@ impl<'de, K, V, S> Visitor<'de> for OrderMapVisitor<K, V, S>
     fn visit_map<A>(self, mut map: A) -> Result<Self::Value, A::Error>
         where A: MapAccess<'de>
     {
-        let mut values = OrderMap::with_capacity_and_hasher(map.size_hint().unwrap_or(0), Default::default());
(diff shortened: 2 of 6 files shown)


src/macros.rs
```diff
@@ -36,3 +36,38 @@ macro_rules! ordermap {
...
+///
+/// // "a" is the first value
+/// assert_eq!(set.iter().next(), Some(&"a"));
+/// # }
+/// ```
+macro_rules! orderset {
+    ($($value:expr,)+) => { orderset!($($value),+) };
+    ($($value:expr),*) => {
+        {
+            let _cap = ordermap!(@count $($value),*);
```
bluss: I'd copy the count part of the macro to this place. It's the only way to support #[macro_use(orderset)] which I'd like to do whenever possible.

src/serde.rs
```diff
@@ -65,3 +65,59 @@ impl<'de, K, V, S> Deserialize<'de> for OrderMap<K, V, S>
...
+    type Value = OrderSet<T, S>;
+
+    fn expecting(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(formatter, "a set")
+    }
+
+    fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+        where A: SeqAccess<'de>
+    {
+        let mut values = OrderSet::with_capacity_and_hasher(seq.size_hint().unwrap_or(0), Default::default());
```
bluss: `S::default()` is the neater style but I see it's just copying the existing code, so that's understandable.

src/set.rs
```diff
@@ -0,0 +1,1114 @@
...
+/// The interface is closely compatible with the standard `HashSet`, but also
+/// has additional features.
+///
+/// # Order
+///
+/// The values have a consistent order that is determined by the sequence of
+/// insertion and removal calls on the set. The order does not depend on the
+/// values or the hash function at all.
+///
+/// All iterators traverse the set in *the order*.
```
bluss: We need a comment on the order in the set operation iterators

src/set.rs
```diff
@@ -0,0 +1,1114 @@
...
+
+    /// Return an iterator over the values of the set, in their order
+    pub fn iter(&self) -> Iter<T> {
+        Iter {
+            iter: self.map.keys()
+        }
+    }
+
+    /// Return an iterator over the values that are in `self` but not `other`.
+    pub fn difference<'a, S2>(&'a self, other: &'a OrderSet<T, S2>) -> Difference<'a, T, S2>
```
bluss: We need a comment on the order, not sure if there's anything interesting we can say. If we prefer to not guarantee anything that's ok. Difference, in particular, seems like a simple example of guaranteeing the order -- same as self, whenever an element appears.

src/set.rs
```diff
@@ -0,0 +1,1114 @@
...
+    }
+
+    fn last(mut self) -> Option<Self::Item> {
+        self.next_back()
+    }
+
+    fn collect<C>(self) -> C
+        where C: FromIterator<Self::Item>
+    {
+        self.iter.map(|(x, ())| x).collect()
```
bluss: For this to reach its goal -- delegate down to the sliceiterator.map(f).collect() -- we need to not go through the OrderMap iterator here, right? The reason we delegate down to the libcore iterators is to get their TrustedLen etc optimizations for collect.

bluss: The other iterator methods should be free of this problem.

cuviper: Ah, I didn't really understand *why* you were overriding `collect()` at all, but it makes sense to leverage the unstable traits this way.  I can make this use `self.iter.iter.map(...).collect()` -- although once we're digging into private fields, maybe it would be better for `set` to just create and use the underlying vec/slice iterator directly, avoiding the doubled `map(..)` in the other methods.

I was *mostly* trying to avoid having the set use private features of the map, but maybe that's not so helpful.  I was somewhat planning for the possibility that you might rather publish this as a separate `orderset` crate.

src/set.rs
cuviper: I think it's fine to guarantee order on all of them.  The tests I wrote do make sure of this, since they compare the iterators directly against range iterators.

src/set.rs
bluss: I think it is fine to ship in the same crate, they are closely related.

src/macros.rs
cuviper: Done.

src/serde.rs
cuviper: I updated both cases.

src/set.rs
cuviper: Done.

src/set.rs
cuviper: I've documented them all.

src/set.rs
cuviper: Done.

cuviper:
I just moved `iterator_methods!` to a common place too, so that can be reused by the `set` iterators.

bluss:
Thanks a lot!

lucab:
I'd like to point out that the behavior of the `OrderSet::insert()` here is actually an "upsert": if the element already exist, its relative order in the set will be updated. I'm not sure if this is intentional design.

If possible, I'd like to:
 * update `insert` to avoid changing order on re-insertion, similarly to other languages/libraries (see top note for [Java LinkedHashSet](https://docs.oracle.com/javase/7/docs/api/java/util/LinkedHashSet.html))
 * have a dedicate `upsert` (or similar name) with the current behavior
 * explicitly document those two different behaviors

Please note that I'm not arguing which of the two designs is better, as I have (different) use cases for both of them.

cuviper:
@lucab 
> I'd like to point out that the behavior of the OrderSet::insert() here is actually an "upsert": if the element already exist, its relative order in the set will be updated. I'm not sure if this is intentional design.

That's not my intention, and doesn't seem to be the case:

```rust
    let mut set: OrderSet<_> = (0..10).collect();
    println!("{:?}", set);
    set.insert(3);
    println!("{:?}", set);
```

That doesn't perturb the order AFAICS.

lucab:
@cuviper you are right, my bad. I misread `OrderMap::insert()` implementation and didn't check with a running example. So minor remaining questions related to my comment above:
 * is this guaranteed/documented somewhere?
 * can we have an explicit `OrderSet::upsert()` for cases where I need to update the order on re-insertion?

For both cases I can happily provide docs/tests/code, depending on the answer.

cuviper:
- IMO this is fine to guarantee.  I don't think this aspect of update behavior is documented on either the map or the set, but it's actually the easier way to implement it anyway.
- An `upsert` (which I would expand as either `update_insert` or `insert_update`) has the same basic problem as `remove` -- what do you do with any following elements?  Should it behave like `swap_remove` and move the last element into the hole?  Should it do an O(n) shift like #15?

cuviper:
(Huh, I didn't realize `upsert` was already a common term in databases.  Maybe that's not too bad then.)

lucab:
Yes, name suggestion comes from https://wiki.postgresql.org/wiki/UPSERT. I'm not attached to it, but at the same time it already exists for this exact behavior.

Regarding removal-then-insert, as a user I'd pretty much want the behavior of #15 even if it could be very costly. Pointing that out in doc will probably give me (as a library user) a clear hint that I should try a different design instead first, and fallback to `upsert` only as a last resort.

bluss:
@lucab yes, if you want strict insertion order, then this is not the right data structure.

PR #47, opened 2017-11-30
WIP: Add views of OrderMap

cuviper:
`OrderMap::view() -> OrderMapView` and `view_mut() -> OrderMapViewMut`
provide access to items in the map, but no ability to add/remove items.
These views can be sub-divided with `split_at` etc., which may be useful
to enable parallelism.

### DIFF
--- a/src/view_mut.rs
+++ b/src/view_mut.rs
@@ -1,0 +1,478 @@
+use super::*;
+
+/// A mutable view of an `OrderMap`.
+///
+/// Mutable views are not just mutable references into the map --
+/// they can also be split into smaller views!
+pub struct OrderMapViewMut<'a, K: 'a, V: 'a, S: 'a = RandomState> {
+    offset: usize,
+    mask: usize,
+    indices: &'a [Pos],
+    entries: &'a mut [Bucket<K, V>],
+    hash_builder: &'a S,
+    size_class_is_64bit: bool,
+}
+
+impl<'a, K, V, S> fmt::Debug for OrderMapViewMut<'a, K, V, S>
+    where K: fmt::Debug + Hash + Eq,
+          V: fmt::Debug,
+          S: BuildHasher,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_map().entries(self.iter()).finish()
+    }
+}
+
+impl<'a, K, V, S> OrderMapViewMut<'a, K, V, S> {
+    /// Creates a read-only view of a map.
+    pub fn new(map: &mut OrderMap<K, V, S>) -> OrderMapViewMut<K, V, S> {
+        OrderMapViewMut {
+            offset: 0,
+            mask: map.mask,
+            indices: &map.indices,
+            hash_builder: &map.hash_builder,
+            size_class_is_64bit: map.size_class_is_64bit(),
+            entries: &mut map.entries,
+        }
+    }
+
+    /// Return the number of key-value pairs in the map view.
+    ///
+    /// Computes in **O(1)** time.
+    pub fn len(&self) -> usize { self.entries.len() }
+
+    /// Returns true if the map view contains no elements.
+    ///
+    /// Computes in **O(1)** time.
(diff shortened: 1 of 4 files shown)


cuviper:
Views were suggested here: https://github.com/bluss/ordermap/pull/45#issuecomment-346435447

The two views have different contents, so I thought I should explain why, and we can decide how to proceed.

`OrderMapView` is just a reference to the original map and range indices.  Everything is simple.

`OrderMapViewMut` has to break it down, because it needs exclusive non-overlapping slices of `entries` in order to be mutable.  One notable difference this causes is that key lookup is less efficient.  If the hash gets an index from `pos` that's not in our range, we have to just assume that the key at that remote entry would not be equal, and thus we must continue the linear probing as if it were a hash collision.  I'm not sure how much this will matter to performance in practice.

This setup also means that `OrderMapViewMut` can't degrade to an `OrderMapView`, which makes `split_at` awkward -- it ought to take `&self` and return immutable `OrderMapView`s.

One idea is that `OrderMapView` could have the same fields broken out, but hang on to as much of `entries` as possible -- so it would need the `offset` *and* the `start`..`end` indexes (end offset is implicit in `entries.len()`).  When created from a map, it would keep referencing the entire `entries` (`offset = 0`) with splits only affecting `start` and `end`.  When created from an `OrderMapViewMut`, it would have that reduced view of `entries`.  This way `OrderMapView` keeps as much as it can for key lookup comparisons, but would still work with a limited view too.

cuviper:
The other thing I wanted to point out is that sadly I don't think it's possible to allow `Index` with ranges.  That has to return a reference, `&Self::Output`, and we don't have a nice `[T]` or `str` type we can use for `Output` like this.

bluss:
Hey, I'm thrilled! Just love to see an idea tried out and come to life

bluss:
Yes the Rust limitations for views is something I know well from ndarray :smile: I also have a preconceived notion from ndarray's view design: it would use [`.split_at(self, index) -> (Self, Self)`](https://docs.rs/ndarray/0.10.13/ndarray/type.ArrayViewMut.html#method.split_at) for both the view types. You could downgrade explicitly with `.view(&self) -> View` and reborrow with `.view_mut(&mut self) -> ViewMut` (but it is never going to equal the slice exactly.).

cuviper:
OK, splitting by value works.

src/view_mut.rs
```diff
@@ -0,0 +1,478 @@
...
+              Sz: Size,
+    {
+        debug_assert!(self.len() > 0);
+        let mut probe = desired_pos(self.mask, hash);
+        let mut dist = 0;
+        probe_loop!(probe < self.indices.len(), {
+            if let Some((i, hash_proxy)) = self.indices[probe].resolve::<Sz>() {
+                // Unfortunately, we can only look at indexes within our slice
+                // of entries, which might mean we'll keep probing farther than
+                // it would be strictly necessary.
```
bluss: We can remedy this (often); by asking for the short hash that is stored in the indices; it is accessible when usize is 64-bit and the entry capacity fits in 32 bits. It requires breaking down `get_short_hash` and using something like this:

```rust
/// Get the hash from the indices if possible.
fn try_short_hash<K, V>(&self, index: usize) -> Option<ShortHash<Sz>> {
    if Sz::is_64_bit() {
        None
    } else {
        Some(ShortHash(self.0, PhantomData))
    }
}
```

cuviper:
I haven't thought about it in a while.  Is this a feature you would like?
I should revisit the Rayon PR #45 too...

bluss:
@RReverser I haven't had much time for github, that goes up and down sometimes.

@cuviper Right, so how do we decide if this is actually useful or just a ridiculously cool idea :-)  I'd say we want this, thanks for all the implementation work but if you want to you can leave it for me to finish it, that's fine with me.  (These kinds of view definitions in Rust, they don't feel entirely perfect in general, not sure if we are missing language features or what we are missing to really express this in a better way; having multiple kinds of equivalent read-only IndexMap-ish types then later demands traits that treats the two as equal and so on..)

cuviper:
#177 did this better.

PR #50, opened 2017-12-01
Tweak README

hcpl:
- Fix typos
- Change wordings
- Maintain a certain degree of consistency in punctuation marks, verbatim style usage, methods syntax and username syntax when making references

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,17 +17,17 @@ __ https://docs.rs/ordermap/
 Background
 ==========
 
-This was inpired by Python 3.6's new dict implementation (which remembers
+This was inspired by Python 3.6's new dict implementation (which remembers
 the insertion order and is fast to iterate, and is compact in memory).
 
 Some of those features were translated to Rust, and some were not. The result
-was ordermap, a hash table that is
+was ordermap, a hash table that has following properties:
 
-- Order is **independent of hash function** and hash values of keys
-- Fast to iterate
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -17,56 +17,56 @@ __ https://docs.rs/ordermap/
...
   second, dense, the key-value pairs.
 
 - Iteration is very fast since it is on the dense key-values.
 - Removal is fast since it moves memory areas only in the first vector,
-  and uses a single swap in the second vector
+  and uses a single swap in the second vector.
 - Lookup is fast-ish because the hashes and indices are densely stored.
   Lookup also is slow-ish since hashes and key-value pairs are stored in
-  separate places. (Visible when cpu caches size is limiting.)
+  separate places. (Visible when cpu caches size is limiting).
```
bluss: It would be silly of me to disagree but.. shouldn't this not change? The whole sentence is inside the ()'s.

hcpl: I actually wanted something like "Sentence (clarification)." but now I think it's better to leave it as it was before because there are already 2 sentences out of parentheses.

hcpl: Force-pushed the change.

hcpl:
It's nothing. I just wanted to have to some fun while practicing to use `ordermap` :laughing:

bluss:
Thanks!

PR #51, opened 2017-12-02
set: specify re-insertion behavior

lucab:
This commit specifies current re-insertion behavior, which does not
perturb insertion order, making such guarantee explicit in docstrings.
It also introduces a duplicate-insert test to ensure the behavior
won't have regressions.

### DIFF
--- a/src/set.rs
+++ b/src/set.rs
@@ -25,7 +25,9 @@ type Bucket<T> = super::Bucket<T, ()>;
 ///
 /// The values have a consistent order that is determined by the sequence of
 /// insertion and removal calls on the set. The order does not depend on the
-/// values or the hash function at all.
+/// values or the hash function at all. Note that insertion order and value
+/// are not affected if a re-insertion is attempted once an element is
+/// already present.
 ///
 /// All iterators traverse the set *in order*.  Set operation iterators like
 /// `union` produce a concatenated order, as do their matching "bitwise"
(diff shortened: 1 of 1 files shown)


bluss:
I'll re-check the ordermap code to be sure, but we should also specify if the new or or old key is stored in the given location. Plain Rust HashMap's `.insert(k, v)` will replace an an existing value with `v` but not use the `k`. It's how that one is, and `HashSet` follows that (using only keys).

bluss:
Checked: yes, that's how it is.

1. Preexisting key-value pair is called k1-v1 (if any)
2. On `map.insert(k2, v2)` where `k2` is equivalent to `k1`:
  a. If no key-value pair k1-v1 existed in the map, k2-v2 are stored after insert in the map. (No values dropped)
  b. If k1-v1 already existed in the map, k1-v2 is stored after insert in the map (k2 and v1 are dropped).

The OrderSet works the same as a map -- every element stored in an OrderSet is like a **key** in a map.

lucab:
Good. So, should I leave it as-is or turn it into an explicit "without re-inserting the value nor altering its insertion order"?

src/set.rs
```diff
@@ -145,10 +146,11 @@ impl<T, S> OrderSet<T, S>
         self.map.reserve(additional);
     }
 
-    /// Insert the value into the map.
+    /// Insert the value into the set.
     ///
-    /// If the value did already exist in the set, returns `false`;
-    /// otherwise, return `true`.
+    /// If the value did already exist in the set, return `false`
+    /// without altering its insertion order; otherwise, insert it and
```
bluss: Please mention that the old value remains and that the new is not used/not inserted.

Maybe just -- *if an equivalent value already existed in the set, it remains and the method returns false.*  This covers both parts

bluss: Ok, maybe mention something about order. It remains in the set, in its place in the order. If you want to.

lucab:
Rebased, PTAL.

bluss:
Thanks a lot!

PR #56, opened 2018-01-02
Make .retain() in-order and preserve-order

bluss:
- Having a retain that keeps the order is very useful
- This is slower across the board but especially if we retain most of the map; the swap_remove version (unordered version) just moves much less values around, even less than a regular `Vec::retain`.
- This PR uses a trick (tombstones in self.indices!) for the duration of the retain method only

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -201,6 +201,21 @@ impl Pos {
             }
         }
     }
+
+    /// Like resolve, but the Pos **must** be non-none. Return its index.
+    #[inline]
+    fn resolve_existing_index<Sz>(&self) -> usize 
+        where Sz: Size
+    {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -959,13 +978,121 @@ impl<K, V, S> OrderMap<K, V, S>
...
+                        self.indices[best_free_pos] = self.indices[probe];
+                        self.indices[probe] = Pos::none();
+                        prev_empty = improvement_shift;
+
+                        // manual probe increment:
+                        // the only way to `continue` inside probe_loop!()
+                        probe += 1;
+                        continue;
+                    }
+                    prev_empty = 0;
```
bluss: wait a minute.. this continue etc sequence is useless

PR #61, opened 2018-02-01
Docs & Relnote for ordermap 0.4.0 (the last release series under this name)

bluss:
The idea is to introduce deprecations and IndexMap type aliases in 0.4.1

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -97,6 +97,23 @@ Ideas that we already did
 Recent Changes
 ==============
 
+- 0.4.0
+
+  - This is the last release series for this ``ordermap`` under that name,
+    because the crate is **going to be renamed** to ``indexmap`` (with types
+    ``IndexMap``, ``IndexSet``) and no change in functionality!
+  - The map and its associated structs moved into the ``map`` submodule of the
+    crate, so that the map and set are symmetric
+
+    + The iterators, ``Entry`` and other structs are now under ``ordermap::map::``
+
+  - Internally refactored ``OrderMap<K, V, S>`` so that all the main algorithms
+    (insertion, lookup, removal etc) that don't use the ``S`` parameter (the
+    hasher) are compiled without depending on ``S``, which reduces generics bloat.
+
+  - ``Entry<K, V>`` no longer has a type parameter ``S``, which is just like
+    the standard ``HashMap``'s entry.
+
 - 0.3.5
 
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -97,6 +97,23 @@ Ideas that we already did
...
+    because the crate is **going to be renamed** to ``indexmap`` (with types
+    ``IndexMap``, ``IndexSet``) and no change in functionality!
+  - The map and its associated structs moved into the ``map`` submodule of the
+    crate, so that the map and set are symmetric
+
+    + The iterators, ``Entry`` and other structs are now under ``ordermap::map::``
+
+  - Internally refactored ``OrderMap<K, V, S>`` so that all the main algorithms
+    (insertion, lookup, removal etc) that don't use the ``S`` parameter (the
+    hasher), are compiled without depending on ``S``.
```
cuviper: Does this have any user-visible impact?  I think it *could* if all such methods were in `impl` without `S: BuildHasher`, as some methods already are, but I don't know if this is very meaningful.  Besides that, I guess the benefit is reduced monomorphization for these implementations?

bluss: Yep, it's intended to reduce generics bloat, even if I guess it might not make the difference in most use cases.

`Entry<K, V, S>` -> `Entry<K, V>` is the user visible thing I can think of (and the component structs of entries). There should be no other visible changes. The refactoring was surprisingly smooth, but I suppose the diff is not so easy to read.  ( `git diff --stat -p 324caad..b159fad` )

That reminds me, I didn't visit OrderSet at all for that refactoring, but I think most of its unique features (like subset / difference) depend on the hasher.

bluss: There's even the closure to `Bucket::key_ref` accessors (and similar) refactoring in there. The main motivation was to reuse the same function instead of having a separate identical closure for each iterator method in the same iterator.

cuviper: Yeah, most of the new set stuff uses `contains(item)`.

Maybe just add a final clause here: "without depending on `S`, reducing generics bloat."

bluss: amended with that, thanks!

bluss:
Since it's a habit I might push & merge easy PRs without warning. Just say if it doesn't work.

cuviper:
No worries.  I'm watching the repo, so I'll be sure to complain if I see something fishy. 😉

Issue #62, opened 2018-02-03
Minimum Rust Version


bluss:
Petgraph uses Ordermap 0.3.x and uses a minimum Rust version of 1.12 (and it compiles), in this repo we test from Rust 1.18 in travis. So I need to look at docs & travis test for that in ordermap. More relevant for 1.0-ing maybe. At the point of an 1.x release series, it's no longer reasonable to freeze the minimum Rust version, so we'd have to consider letting it upgrade (slowly).

bluss:
We've made a policy like this for 1.x, so the issue is complete. See crate docs for info.

PR #63, opened 2018-02-03
Just Travis: What's the real version baseline?

bluss:
Related to #62


bluss:
Oh right, I put in `pub(crate)` in several places now ☹️

bluss:
Ok, so the current baseline is Rust 1.18 due to the pub(crate) syntax.

PR #64, opened 2018-02-12
Add deprecation notice to OrderMap/Set to prepare for rename to indexmap

bluss:
It's possible to put the deprecation on the crate itself, but then users
get a ridiculous onslaught of messages for each item (including methods)
used from the crate. So we put the notice on OrderMap and OrderSet
themselves.


bluss:
This is for ordermap 0.4.1, which should be the final ordermap release under this name. (Then comes indexmap)

PR #65, opened 2018-02-12
Rename OrderMap to IndexMap; rename crate to indexmap (and OrderSet to IndexSet)

### LINKED ISSUE #14: Change name?
> original-insertion-order-preserving removal would want to be implemented with tombstones, but I've decided this is going to be a separately named hash map. So: fork time.

Just an idle thought -- OrderMap was a good name for something that does guarantee insertion-order-preserved for all operations, so have a new name for this new thing which does not make that guarantee.

Is there a FastMap ? :-)

bluss:
- `OrderMap` becomes `IndexMap` with no change in functionality
- Releasing version 1.0 of the crate becomes possible shortly after the rename
- This is the big rename; see discussion and rationale in issue #14

Plan:

1. Release ordermap 0.4.1 with a deprecation message pointing to indexmap; in PR #64
2. Release indexmap 0.4.1 with the rename (and deprecated type aliases in place for OrderMap/Set) (this PR)
3. Release indexmap 1.0 with deprecated items removed. (later)

Fixes #14

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,9 +1,9 @@
 [package]
-name = "ordermap"
+name = "indexmap"
 version = "0.4.1"
 authors = ["bluss"]
-documentation = "https://docs.rs/ordermap/"
-repository = "https://github.com/bluss/ordermap"
+documentation = "https://docs.rs/indexmap/"
+repository = "https://github.com/bluss/indexmap"
 license = "Apache-2.0/MIT"
 description = "A hash table with consistent order and fast iteration."
 
--- a/README.rst
+++ b/README.rst
@@ -3,16 +3,22 @@ Awesome hash table implementation in just Rust (stable, no unsafe code).
 
 Please read the `API documentation here`__
 
-__ https://docs.rs/ordermap/
+__ https://docs.rs/indexmap/
 
 |build_status|_ |crates|_
 
-.. |crates| image:: https://img.shields.io/crates/v/ordermap.svg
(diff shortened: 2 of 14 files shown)


bluss:
PR #64 needs to be merged & released before this one.

cuviper:
> Release indexmap 0.4.1 with the rename (and deprecated type aliases in place for OrderMap/Set) (this PR)

Why not put the deprecated type aliases in `ordermap`?  Then `indexmap` can head straight for 1.0, or at least never pollute itself with deprecations.

bluss:
I guess my idea would be that each step is just one change, like ordermap => indexmap with equivalent version number. I'm not sure if that's valuable in any way, to have that obsolete version correspond, once indexmap 1.0 is released. I think it can still have some value.

Having a deprecated `OrderMap` and a type alias `IndexMap` in crate ordermap 0.4.1 would be possible but it is incomplete, since the user needs to update their dependency to point to indexmap anyway. The original plan was though to have both of these, deprecated parts on both sides of the rename.

I am thinking about the Rust version requirement (Rust version requirement was never documented anyway, travis was using 1.18 anyway, but apparently petgraph was using it from its Rust 1.12). Working around not having pub(crate) is about equally annoying though.

bluss:
It doesn't have to be the perfect strategy, it's just a set of small steps that leads us to indexmap 1.0. So I won't try to work on the perfect strategy here.

PR #66, opened 2018-02-19
Fix new() not working with custom hashers

RReverser:


cuviper:
This design is following the lead of the standard `HashMap`.  It seems like a design wart, but then note that you now have a type-inference failure revealed in CI:

```
error[E0282]: type annotations needed
  --> tests/tests.rs:9:13
   |
9  |       let m = indexmap! {
   |  _________-___^
   | |         |
   | |         consider giving `m` a type
10 | |         1 => 2,
11 | |         7 => 1,
12 | |         2 => 2,
13 | |         3 => 3,
14 | |     };
   | |_____^ cannot infer type for `S`
   |
   = note: this error originates in a macro outside of the current crate
```

I expect you could also make the macros set `S` explicitly, but take it as a warning that it may break outside users in a similar way.

If we decide to change anyway, it should be done for `IndexSet` too.

RReverser:
>  It seems like a design wart

Yeah it did feel like gotcha when `OrderMap::new()` didn't work when using with `fnv` hasher, but later found out that `OrderMap::default()` does work.

> I expect you could also make the macros set S explicitly, but take it as a warning that it may break outside users in a similar way.

That's a good case, I should've ran all the tests locally. Not sure what's the best way forward here... On one hand, yeah, it's easy to set default type param just in macro, on another 1) there might be many more similar cases and 2) actually enabling macro to work with custom hashers could be also nice...

> If we decide to change anyway, it should be done for IndexSet too.

Sure, just forgot about it.

RReverser:
TBH I'm a bit surprised that Rust itself doesn't choose type param default for static methods in that case. But given how many cases this breaks, I guess it's a no-go for now...

bluss:
Thanks I agree with closing this since we are mimicking HashMap so closely. A brief hint about `IndexMap::default` in the doc comment would be good improvement to `new`.

PR #69, opened 2018-03-04
Change OccupiedEntry::insert to use &mut self

### LINKED ISSUE #68: Signature of OccupiedEntry::insert
- [OccupiedEntry::insert for HashMap](https://doc.rust-lang.org/std/collections/hash_map/struct.OccupiedEntry.html#method.insert)
- [OccupiedEntry::insert for BTreeMap](https://doc.rust-lang.org/std/collections/btree_map/struct.OccupiedEntry.html#method.insert)
- [OccupiedEntry::insert for LInkedHashMap](https://docs.rs/linked-hash-map/0.5/linked_hash_map/struct.OccupiedEntry.html#method.insert)
- [OccupiedEntry::insert for IndexMap](https://docs.rs/indexmap/0.4/indexmap/map/struct.OccupiedEntry.html#method.insert)

The first three all take `&mut self` while indexmap takes `self` by value. Is this intentional? I found this makes it complicated for a library to multiplex across different choices of underlying map type: https://github.com/serde-rs/json/issues/418.

My workaround is:

```rust
mem::replace(occupied_entry.get_mut(), value)
```

bluss:
This fixes an inconsistency compared with HashMap; this way we use the
same signature for our OccupiedEntry::insert.

Fixes #68 

Thanks @dtolnay

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -588,8 +588,9 @@ impl<'a, K, V> OccupiedEntry<'a, K, V> {
         &mut self.map.entries[self.index].value
     }
 
-    pub fn insert(self, value: V) -> V {
-        replace(&mut self.into_mut(), value)
+    /// Sets the value of the entry to `value`, and returns the entry's old value.
+    pub fn insert(&mut self, value: V) -> V {
+        replace(self.get_mut(), value)
     }
 
     pub fn remove(self) -> V {


bluss:
With this, we are go for 1.0 right?

cuviper:
I think so!  I did another sweep, and didn't notice anything else different from `HashMap` and `HashSet`.  Just missing `shrink_to_fit` as #19 already notes, and I didn't fully check that all trait implementations match, but that can be extended later.

PR #70, opened 2018-03-11
Document minimum rust version and versioning policy

bluss:
We now have indexmap 1.0!

We need to 

1. Document our version requirement (so that others can understand it and use it)
2. Make a sustainable plan for the version requirement in 1.0

I believe we can have a very long lived indexmap 1.x release series if we are careful keepers of the crate. We do then, however, have to deal with a changing Rust landscape, and whatever it will bring. I think it's best to be up front about this: I don't think we want or can stick to a Rust 1.18 minimum requirement forever. I don't either think that we should rush to new versions (I already think the Rust 1.18 requirement is a bit high for the present time, but that can be fixed backwards compatibly..)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,6 +10,15 @@
 //!
 //! [`IndexMap`]: map/struct.IndexMap.html
 //! [`IndexSet`]: set/struct.IndexSet.html
+//!
+//!
+//! # Minimum Rust Version
+//!
+//! This version of indexmap requires Rust 1.18.
+//!
+//! The indexmap 1.0 release series will use a carefully considered version
+//! upgrade policy, where in a later 1.x version, we will raise the minimum
+//! required Rust version.
 
 #[macro_use]
 mod macros;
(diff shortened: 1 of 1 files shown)


bluss:
cc @cuviper Tricky stuff. I think it's good to be up front about this.

cuviper:
Yeah... I'm very cautious about increasing rustc requirements, but I wish the tools handled this better.  Bumping the crate's minor version for this is pretty reasonable, at least.

src/lib.rs
```diff
@@ -10,6 +10,15 @@
 //!
 //! [`IndexMap`]: map/struct.IndexMap.html
 //! [`IndexSet`]: set/struct.IndexSet.html
+//!
+//!
+//! # Minimum Rust Version
+//!
+//! This version of indexmap requires Rust 1.18.
+//!
+//! The indexmap 1.0 release series will use a carefully considered version
```
bluss: This can be fixed to say "indexmap 1.x release series"

bluss:
Ok, let's go with this. It's important to have both here, document the supported version and the policy around it. Long live 1.x.

PR #74, opened 2018-06-15
docs: improve badges and synopsis

lucab:
This re-arranges the first section of README, so that the short synopsis
provides a quick and complete overview of the crate, and further info
is linked through badges.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,11 +1,7 @@
+indexmap
+========
 
-Awesome hash table implementation in just Rust (stable, no unsafe code).
-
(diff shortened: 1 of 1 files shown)


cuviper:
Works for me -- maybe we should link the readme in `Cargo.toml` too?  Does crates.io understand `rst`?

lucab:
It doesn't. It expects CommonMark syntax. See https://github.com/rust-lang/crates.io/pull/869.

lucab:
Bump on this.

PR #77, opened 2018-08-03
Add `or_default` method to `map::Entry`

mexus:
P for Parity :) In response to https://github.com/bluss/indexmap/pull/76#issuecomment-410063622


mexus:
I didn't know you are looking for the parity with the std's hash_map::Entry, otherwise I would have done it from the very beginning :)

cuviper:
Oh, there actually is #19 tracking this, and `or_default` is on the list.  We should update that.

Anyway, thanks!

PR #79, opened 2018-09-10
Update quickcheck to 0.7

ignatenkobrain:


ignatenkobrain:
Ignore me, this needs more work because of porting to new rand.

PR #85, opened 2018-11-24
Update quickcheck to 0.7 and rand to 0.5

bluss:
Potentially upgrade these deps. Rand will require Rust 1.22, but it's only used for testing, so it could be moved to a separate testing crate that doesn't run for earlier versions.


cuviper:
I'm OK with just doing a build on older compilers, leaving tests to newer.  Might be good to have an explicit version where we expect tests to fully work though, rather than leaving it to just `stable` and on.

While we're at it, rand 0.6 is out now.

mwillsey:
I think this is superseded by #97, which was just merged.

PR #87, opened 2019-01-02
Implement IntoDeserializer for IndexMap/IndexSet

### LINKED ISSUE #86: Implement serde::IntoDeserializer for IndexMap/IndexSet
It must be trivial because serde already has generic [`MapDeserializer`]: https://docs.serde.rs/src/serde/de/value.rs.html#1254-1266

[`MapDeserializer`]: https://docs.serde.rs/serde/de/value/struct.MapDeserializer.html

Mingun:
Fixes #86

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -116,6 +116,8 @@ Recent Changes
   - Implemented ``Debug`` for ``map::{Entry, IntoIter, Iter, Keys, Values}`` and
     ``set::{Difference, Intersection, IntoIter, Iter, SymmetricDifference, Union}``
 
+  - Serde trait ``IntoDeserializer`` are implemented for ``IndexMap`` and ``IndexSet``.
+
 - 1.0.2
 
   - The new methods ``IndexMap::insert_full`` and ``IndexSet::insert_full`` are
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -2,7 +2,8 @@
 extern crate serde;
 
 use self::serde::ser::{Serialize, Serializer, SerializeMap, SerializeSeq};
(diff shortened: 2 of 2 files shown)


bluss:
Thanks. I'd generally prefer avoiding bumping versions in feature PRs, best leave the versions to the mantainers. That said, adding to the changelog is great, so we still risk merge conflicts there.

bluss:
I see that we don't test serde on the minimum version in travis. Can we figure out the earliest possible version we can test serde on?

cuviper:
I tried this PR with serde_test reverted back to 1.0.5, and used `cargo update -Zminimal-versions`. Everything was fine except we need fnv 1.0.5 for `FnvBuildHasher`.

Mingun:
The reason for which I bump version of the `serde_test` in avoiding a lot of copies of the same libraries taking the place on the developer machine working with many projects. When all dependences use the latest versions, it is less probable. Besides, I do not see any reason not to bump the versions of the test dependencies, especially, if all changes are compatible. With high probability, there something is fixed. However, if you against such approach, I can revert this change

cuviper:
Please do revert the bumped dependency, then we should merge this.

Remember that Cargo uses semver compatibility to resolve dependencies, so "avoiding a lot of copies" is not really an issue. Having `serde_test = "1.0.5"` is equivalent to `"^1.0.5"` or `">=1.0.5 <2.0.0"`. It would only pin an exact dependency with `serde_test = "=1.0.5"`.

Mingun:
Done.

cuviper:
Thanks!

PR #88, opened 2019-01-02
Implementation of Clone, Debug and documentation for structures

Mingun:
Resolves two cases from #19. I add only trivial implementations -- for example, for `Drain` implementing of `Debug` not trivial, so I do not add it

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "indexmap"
-version = "1.0.2"
+version = "1.1.0"
 authors = [
 "bluss",
 "Josh Stone <cuviper@gmail.com>"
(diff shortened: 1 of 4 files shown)


cuviper:
These all look good to me, thanks!

AFAICS this doesn't affect the minimum supported rustc though, so let's make it 1.0.3 instead of bumping to 1.1. The policy documented in src/lib.rs is to keep consistent for any given 1.x series.

bluss:
I don't mind bumping to 1.1 while keeping the same minimum Rust version. Since these are new features, 1.1 seems more correct.

cuviper:
OK, I don't feel strongly about the version, if you're both happy with 1.1.

Mingun:
ping

cuviper:
I added `Debug` for `IntoIter` using `vec::IntoIter::as_slice()` to access the items. I think we could also do `Drain` and `IterMut` if `vec::Drain` and `slice::IterMut` exposed slices, which I'll propose in `std`. For now, this seems good enough.

Thanks!

PR #92, opened 2019-04-28
Enable use from no_std crates

yaram:
This PR adds the ability to use this crate with `no_std`. It has no breaking changes, using a default `std` feature with an optional `core` feature. Since `HashMap` is only included in `libstd`, it uses the `hashmap_core` crate to replace it when compiling with `no_std`.

It may be cleaner to replace uses of `std` with `core` or `alloc`, rather than creating a fake `std`. Please let me know if you'd prefer that.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,6 +20,34 @@
 //! upgrade policy, where in a later 1.x version, we will raise the minimum
 //! required Rust version.
 
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[cfg(not(feature = "std"))]
+#[macro_use]
+extern crate alloc;
+
+#[cfg(not(feature = "std"))]
+extern crate hashmap_core;
+
+#[cfg(not(feature = "std"))]
+mod std {
+    pub use core::*;
+
+    pub mod vec {
+        pub use alloc::vec::*;
+    }
+
+    pub mod boxed {
+        pub use alloc::boxed::*;
+    }
+
+    pub mod collections {
+        pub mod hash_map {
+            pub use hashmap_core::map::*;
+        }
+    }
+}
+
 #[macro_use]
 mod macros;
 #[cfg(feature = "serde-1")]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ bench = false
 
 [dependencies]
 serde = { version = "1.0", optional = true }
+hashmap_core = { version = "0.1.10", optional = true }
 
 [dev-dependencies]
 itertools = "0.7.0" # 0.8 not compiles on Rust 1.18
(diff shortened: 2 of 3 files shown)


cuviper:
> It has no breaking changes

Unfortunately, this **is** a breaking change, for two reasons:

1. Features must be additive, in that enabling a feature must not constitute a breaking change. Consider if two crates in a dependency both need `indexmap`, but one asks for `no-default-features` (for `no_std`) and the other uses the default with "std". In this case, changing the "std" feature entirely changes the `RandomState` type used for the default `S` parameter, which could break the one that didn't ask for "std".
    - We could work around this by not providing a default `S` at all for `no_std`, which also requires gating the `new` and `with_capacity` methods.

2. Someone could already be depending on `indexmap` with `no-default-features`, expecting this to work with `std`, not needing the newly-stabilized `alloc`. So restricting the current `std` functionality to a new feature flag also constitutes a breaking change, even if it's a default!
    - I ran into this when I tried to implement `no_std` for the `num` crates -- see https://github.com/rust-num/num/issues/297. The eventual solution was to do this in a new semver version, but then use the [semver trick](https://github.com/dtolnay/semver-trick) to have the old version re-export everything from the new with "std" enabled.
    - So we could release `indexmap` 2.0 where "std" is a default feature, and then a new `indexmap` 1.0.z which re-export the std-enabled items from 2.0.

So there are problems here, but I think they can be solved. It's a little annoying to need 2.0 though -- do you have a motivating use case for this? If it's just "because we can", I'd say we shouldn't bother...

> It may be cleaner to replace uses of `std` with `core` or `alloc`, rather than creating a fake `std`. Please let me know if you'd prefer that.

Yeah, style-wise I'd prefer to use `core` directly, and only use cfg `alloc` for `Vec` and `Box` as needed.

src/lib.rs
```diff
@@ -20,6 +20,34 @@
...
+        pub use alloc::vec::*;
+    }
+
+    pub mod boxed {
+        pub use alloc::boxed::*;
+    }
+
+    pub mod collections {
+        pub mod hash_map {
+            pub use hashmap_core::map::*;
```
cuviper: I just realized this changes our default `RandomState` type, normally defined only in `std`, which is also a breaking change.

bluss:
> Someone could already be depending on indexmap with no-default-features, expecting this to work with std, not needing the newly-stabilized alloc.

IMHO we don't need to care about this. Because this crate has no default features to start with, we should not let ourselves be held hostage by those that configure like this. I.e `default-features=false` has been a working but meaningless configuration until we go live with some default feature.

src/lib.rs
bluss: Yes, I think hashmap_core should be dropped from this PR, because it only adds this, which we don't want.

We'd like to not have a default hasher, that should work, and not have IndexMap::new() at all when we are no-std.

cuviper:
Well FWIW, I got called out on default-features for num: https://github.com/rust-num/num/issues/297

You could choose to write off such users, but #95 shows how to do it without features.

bluss:
Ok! I agree, #95 seems good and you've taken it in the right direction for sure. I just didn't understand the build script stuff (because it was not complete).

Issue #93, opened 2019-05-16
A design for single-pass borrowed insert_if_not_there


CAD97:
I have an `IndexSet<String>` that I want to be able to insert or get index from `&str` without allocating needlessly if the value is already in the set.

This is actually possible with one small library addition: `entry_index`. It would return an entry for the value at the index up to and including `.len()`, thus allowing insertion of the value there. It wouldn't even need to be unsafe: having two keys that compare equal in the map is only very bad for domain correctness, but not for memory correctness if I understand correctly.

It could then be used for inserting if not present with something like the following:

```rust
set.get(&key)
    .unwrap_or_else(|| set.entry_index(set.len)
        .insert(key.to_owned())
```

cuviper:
In parity with the `std` sets, there isn't an entry API at all yet. How about this:

```rust
set.get(key).unwrap_or_else(|| {
    let (i, _) = set.insert_full(key.to_owned());
    set.get_index(i).unwrap()
})
```

The `unwrap` is unfortunate, but might just get optimized away in the vector bounds check.

CAD97:
This does need to do a double lookup, and is what I've got written right now. (As `insert` needs to scan for a hash equivalence, which while ~O(1), is not free.) The entry idea is to avoid the double lookup, and would apply equally to maps (I just wanted it on a set). Unfortunately some [other issues](https://github.com/rust-lang/rust/issues/60886) mean that design isn't tenable anyway.

cuviper:
Right, so it seems like you really want a *borrowing* version of `entry` on maps, and then mirror these on sets. That functionality is not really specific to `indexmap`'s design, so I would love to see that API hammered out in `std` first, and then we can mirror it.

I believe `raw_entry_mut` is the current plan:
https://doc.rust-lang.org/nightly/std/collections/struct.HashMap.html#method.raw_entry_mut

cuviper:
Something you may not have realized -- your `entry_index(_).insert(_)` would also count as a second lookup, because entries have two records: the indexed vector and the hash table mapping to indexes. So looking up the index is cheap, but inserting the value would also need to do the raw hash insert.

You can functionally think of `IndexMap` like this:

```rust
struct IndexMap<K, V> {
    map: HashMap<K, usize>,
    vec: Vec<(K, V)>,
}
```

... but optimized to not duplicate key storage.

cuviper:
[...]

PR #95, opened 2019-08-16
no_std support

### LINKED ISSUE #94: Make this crate no_std
Using the new `alloc` crate released with Rust 1.36 it is possible to make the IndexMap independent from std.
It is suggested to make library crates use only the alloc crate in order to be compatible with the `#![no_std]` ecosystem

leo60228:
This bumps the MSRV to 1.36, since it requires the `alloc` crate. Closes #94.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,5 +1,9 @@
+extern crate autocfg;
+
 fn main() {
-    let ac = autocfg::new();
-    ac.emit_has_path("std");
-    ac.emit_has_path("alloc");
+    //let ac = autocfg::new();
+    //ac.emit_sysroot_crate("std");
+    //ac.emit_sysroot_crate("alloc");
+    autocfg::emit("std");
+    //autocfg::emit("alloc");
 }
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,5 @@
+fn main() {
+    let ac = autocfg::new();
+    ac.emit_has_path("std");
+    ac.emit_has_path("alloc");
+}
--- a/src/map.rs
+++ b/src/map.rs
@@ -1,23 +1,27 @@
 //! `IndexMap` is a hash table where the iteration order of the key-value
 //! pairs is independent of the hash values of the keys.
 
-pub use mutable_keys::MutableKeys;
+pub use crate::mutable_keys::MutableKeys;
 
-use std::hash::Hash;
-use std::hash::BuildHasher;
-use std::hash::Hasher;
-use std::iter::FromIterator;
-use std::collections::hash_map::RandomState;
-use std::ops::RangeFull;
+use core::hash::Hash;
+use core::hash::BuildHasher;
+use core::hash::Hasher;
+use core::iter::FromIterator;
+use hashbrown::hash_map::DefaultHashBuilder;
+use core::ops::RangeFull;
 
-use std::cmp::{max, Ordering};
-use std::fmt;
-use std::mem::{replace};
-use std::marker::PhantomData;
+use core::cmp::{max, Ordering};
+use core::fmt;
+use core::mem::{replace};
+use core::marker::PhantomData;
 
-use util::{third, ptrdistance, enumerate};
-use equivalent::Equivalent;
-use {
+use alloc::vec::Vec;
+use alloc::boxed::Box;
+use alloc::vec;
+
+use crate::util::{third, ptrdistance, enumerate};
+use crate::equivalent::Equivalent;
+use crate::{
     Bucket,
     HashValue,
 };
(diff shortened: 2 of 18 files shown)


cuviper:
See also #93 -- but you did it without a feature, which avoids the breaking change there.

I would still rather be more conservative about MSRV though, since 1.36 is very recent. Maybe instead we can make this conditional on the version, detected from a build script? My own `autocfg` crate is useful here, or others like `rustc_version` work too. Then we could be automatically `no_std` for 1.36 and later, while still keeping older support.

src/map.rs
```diff
@@ -260,7 +265,7 @@ impl<Sz> ShortHashProxy<Sz>
 /// assert_eq!(letters.get(&'y'), None);
 /// ```
 #[derive(Clone)]
-pub struct IndexMap<K, V, S = RandomState> {
+pub struct IndexMap<K, V, S = DefaultHashBuilder> {
```
cuviper: Oh, this is a breaking change though, and `RandomState` is only defined in `std`.

leo60228: Is there any scenario in which this causes problems? Type inference should solve it in nearly all cases; I'd be surprised if there's anything that would actually depend on type inference guessing the type of the hasher.

cuviper: It's debatable how likely this is, but if someone was explicit about `S` in one place and assumed the default elsewhere, it would break. Something like:

```rust
type MyMap = IndexMap<MyKey, MyValue, RandomState>; // explicit `S`

fn create_my_map(args...) -> MyMap {
    let mut map = IndexMap::new(); // this uses the default `S`!
    // fill the map
    // ...
    map
}
```

That's a direct breaking change, where something would stop compiling.

There's also the softer semantic change to consider in DOS-resistance, from `std`'s SipHash to `hashbrown`'s choice of aHash. That does still claim to be DOS-resistant, but I don't know enough to evaluate this. If it's really all beneficial, faster and still secure, then aHash should be advocated for `std` itself, and I'd feel better about following suit.

leo60228: aHash only supports x86 and ARM. hashbrown falls back to FxHash on other architectures.

cuviper: Ah, then I'm even more wary of this change.

leo60228: The `siphash` crate provides a `no_std` implementation of SipHash. Should I add a dependency on it?

cuviper: That crate looks abandoned -- I'd be more inclined to roll our own internally. It also doesn't provide a `BuildHasher`, which AIUI needs randomization to be DOS-resistant.

Changing from `RandomState` is still a breaking change at the type level though.

cuviper: A possible way out -- detect the availability of `std` in the build script. If present, we can use `S = RandomState`, otherwise I'd leave it with no default at all. The default-`S` methods `new` and `with_capacity` would also be gated entirely.

So we'd only truly be `no_std` for crates that don't have `std` at all, but I think that's enough. We don't support those targets at all right now, so it's not a breaking change.

leo60228:
Looks like a dependency of rand increased the MSRV upstream.

cuviper:
Ugh, that's annoying. It may be better for us to commit a safe/compatible `Cargo.lock`, which we can remove or update in newer CI runs.

leo60228:
`AutoCfg::probe_path` doesn't work for crates, I'm going to try and submit a PR for `probe_sysroot_crate`/`emit_sysroot_crate`.

bluss:
Looks interesting.

Can this PR formulate the goals and gains that come from this change? The breaking changes and churn require solid motivation. Thanks :slightly_smiling_face:

leo60228:
@cuviper Wait, no, I just forgot that I switched to Rust 2018 in the initial PR...

leo60228:
No idea why the doctests aren't compiling...

build.rs
```diff
@@ -0,0 +1,9 @@
+extern crate autocfg;
+
+fn main() {
+    //let ac = autocfg::new();
+    //ac.emit_sysroot_crate("std");
+    //ac.emit_sysroot_crate("alloc");
+    autocfg::emit("std");
```
bluss: What does this do? Commit logs don't explain it.

bluss:
Also, there is no reason to require Rust 1.36 for alloc — before Rust 1.36, it should still compile using std.

build.rs
cuviper: That would blindly [emit](https://docs.rs/autocfg/0.1.6/autocfg/fn.emit.html) a `std` config value. I suppose they were doing that as a stopgap until the commented crate tests were available -- that's now in autocfg 0.1.6.

PR #97, opened 2019-08-19
Update development dependencies

### LINKED ISSUE #96: 1.0.3 release?
What is needed to cut a 1.0.3 release?

I need the clonable iterators feature introduced in https://github.com/bluss/indexmap/commit/d8e5225a6183886045ded56cc278139ce87d1f17, as it provides parity with HashMap. I'd prefer to not depend on a git revision if possible. 

Is there some way I could help out on preparing a release?

(awesome create btw. I just sped up an application by > 30% by just switching to indexmap!)

mwillsey:
Also update the travis script and the README to indicate that MSRV is
1.18 for builds, and 1.30 for development builds.

Fixes #96.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,7 +14,8 @@
 //!
 //! ## Rust Version
 //!
-//! This version of indexmap requires Rust 1.18 or later.
+//! This version of indexmap requires Rust 1.18 or later, or 1.30+ for
+//! development builds.
 //!
 //! The indexmap 1.x release series will use a carefully considered version
 //! upgrade policy, where in a later 1.x version, we will raise the minimum
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,11 @@ language: rust
 sudo: false
 matrix:
   include:
+    # MSRV is lower for non-dev builds
     - rust: 1.18.0
-      before_script:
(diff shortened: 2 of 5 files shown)


bluss:
Instead of "Addresses #96" please use one of the words that git prefers for closing issues: https://help.github.com/en/articles/closing-issues-using-keywords :)

src/lib.rs
```diff
@@ -14,7 +14,8 @@
 //!
 //! ## Rust Version
 //!
-//! This version of indexmap requires Rust 1.18 or later.
+//! This version of indexmap requires Rust 1.18 or later, or 1.30+ for
+//! development builds.
```
bluss: Nice catch. It's unfortunately reasonable to say it either way - the minimum Rust version is 1.18 if you trust it continues to work and 1.30 if you a crate that actually has tests.

I don't know how silly the discussion is, but I'm a bit uncomfortable with saying that 1.18 is a supported version. But it should be ok, IMO it's not worth dwelling on, at this point.

mwillsey: Fair enough. The travis at least tests what we claim here (builds on 1.18, tests on 1.30).

mwillsey:
Force push added the "fixes #96" to the commit

bluss:
Thanks a lot!

PR #98, opened 2019-08-20
Fix compliation of benches with rand 0.6

bluss:
Also make sure the benches are built by travis (but not run).
And fix one unrelated warning in bench code.


bluss:
oops, I thought this was using bencher for stable benchmarks.. will have to fix

PR #99, opened 2019-08-20
Expand removal options

### LINKED ISSUE #90: Question about ordered remove
I wanted to remove an element without perturbing the order of the remaining elements, and used something like the code below. Is there a better way than this? (In my case it is not really that important, this works just fine for me.)

```rust
type Map = IndexMap<K, V>;
fn remove_ordered(m: &mut Map, key: &K) -> Option<V> {
    let (index, _key, val) = m.swap_remove_full(key)?;
    if index < m.len() {
        let tail = m.len() - index - 1;
        let mut stack = Vec::with_capacity(tail);
        for _ in 0..tail {
            stack.push(m.pop().expect("disappearing elements"));
        }
        let (last_key, last_val) = m.pop().expect("disappearing elements");
        m.extend(stack.into_iter().rev());
        m.insert(last_key, last_val);
    }
    Some(val)
}
```

cuviper:
This aims to make the performance tradeoffs explicit when removing items.

- New `shift_remove` and `shift_take` perform O(n) removal, like `Vec::remove`
- The existing `swap_*` methods already perform like `Vec::swap_remove`
- Ambiguous `remove` and `take` methods, which currently swap, are deprecated in favor of explicit `shift`/`swap` methods.

Fixes #90.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -574,7 +574,7 @@ fn ordermap_merge_shuffle(b: &mut Bencher) {
 }
 
 #[bench]
-fn remove_ordermap_100_000(b: &mut Bencher) {
+fn swap_remove_ordermap_100_000(b: &mut Bencher) {
     let map = OMAP_100K.clone();
     let mut keys = Vec::from_iter(map.keys().cloned());
     let mut rng = SmallRng::from_entropy();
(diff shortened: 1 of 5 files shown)


cuviper:
As we're talking about a new release, I realized that I was also sitting on these changes in a branch...

bluss:
Makes sense. The release is already pushed out, but there can be more releases :slightly_smiling_face: 

What kind of deprecation is intended? My first thought that it would be a "permanently" deprecated method, since one design guideline was drop-in compatibility with HashMap. So you can switch hashmap implementations easily, and with the warning you can realize you have a choice of removal option.

I'm afraid the deprecation message will be annoying for those that don't mind the current behaviour?

bluss:
looks good to me

cuviper:
> What kind of deprecation is intended? My first thought that it would be a "permanently" deprecated method, since one design guideline was drop-in compatibility with HashMap. So you can switch hashmap implementations easily, and with the warning you can realize you have a choice of removal option.

Yes, this makes sense to me. We have a similar situation on rayon [`find`](https://github.com/rayon-rs/rayon/blob/83b67e27f254c5bde039fdc0ea0840f309ce5f73/src/iter/mod.rs#L1739-L1747) with a deprecation note. I think later we added the `#[doc(hidden)]` to avoid clutter.

> I'm afraid the deprecation message will be annoying for those that don't mind the current behaviour?

Maybe so, but it's an easy fix to use `swap_` or `#[allow(deprecated)]`. Do you envision folks wanting to flip back and forth? I'd think that would only be during some initial evaluation.

bluss:
Thank for adding this!  tests/ directory has no mention of shift_remove :sweat_smile: and I'm adding a quickcheck test now for that reason.

bluss:
Indexmap 1.2 is released including this feature :slightly_smiling_face:

PR #101, opened 2019-08-28
Add quickcheck test for shift_remove

bluss:
Follow-up on #99


bluss:
Sure, I'll try to add it to the PR next time - just unsure about how it works, so we'll see then

PR #102, opened 2019-08-28
Clean up some deprecations and clippy lints

cuviper:


bluss:
Seems like an overzealous lint! Nice cleanup though, a bit of a modernization.

PR #104, opened 2019-09-08
Implement index-sliced key-value iterators

### LINKED ISSUE #103: Feature request: `.values_mut_range(idx..)`
I was hacking on the compiler and I wrote:
```rust
                    // We've rejected overlap in each product in the sum.
                    // Now we must account for the possibility that or-pattern is a factor
                    // in a product. A basic case to reject here is `(V1(a) | V2(a), a)`.
                    let last_id = *prod_ids.last().unwrap();
                    bindings.values_mut().skip(len_before).for_each(|val| *val = last_id);
```
which is nice.

Maybe `bindings.values_mut_range(len_before..).for_each(..)` would be nicer?

bluss:
Add iterators that iterate a sliced key-value range (sliced in the
order/index space).

Before merging, we'll look at if it's not better to express using views (from #47).

The main infidelity of the interface is the need to handle
Bound::Exclusive(std::usize::MAX) which does not have an equivalent in
the i..j or i..=j and slices can't handle it natively.

The scaffolding in src/util to create the .slice() and .slice_mut()
methods is somewhat verbose, but it allows us to write neat code where
it is used.

Fixes #103

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,6 +1,9 @@
 
 use std::iter::Enumerate;
 use std::mem::size_of;
+use std::ops::Bound;
+use std::ops::RangeBounds;
+use std::ops::{Range, RangeFrom, RangeInclusive};
 
 pub fn third<A, B, C>(t: (A, B, C)) -> C { t.2 }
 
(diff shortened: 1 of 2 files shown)


src/util.rs
```diff
@@ -15,3 +18,163 @@ pub fn ptrdistance<T>(a: *const T, b: *const T) -> usize {
...
+    //
+    // Range exclusive from usize max to Included/Excluded usize value is
+    // an ill-formed range because it does not have start <= end, but we
+    // don't make the difference here.
+    //
+    // start at i + 1 but panic for oob on overflow
+    let (iplus1, overflow) = i.overflowing_add(1);
+    if overflow {
+        panic!(concat!("Range out of bounds: exclusive from ",
+                stringify!(std::usize::MAX)));
```
cuviper: That `stringify!` prints the tokens, not the constant value -- `std :: usize :: MAX`. You could use a runtime formatting string for `panic!` -- this is a cold case anyway. Or just say, "exclusive from usize::MAX".

bluss: printing the tokens was intended, but the whitespace isn't pretty. I guess I can just use the simple option of writing usize::MAX :)

bluss:
Converting to Range makes sense, but I have been stuck on this thought that this way would be the optimal way to do it for removing as many bounds checks as possible. However, (now looking) the std slicing code does not pay attention to making the tightest possible code for RangeInclusive's bounds checking in slicing, so it is rather a moot point right now.

PR #106, opened 2019-10-01
Support building without libstd

### LINKED ISSUE #94: Make this crate no_std

bluss:
Recast version of #95 by @leo60228

1. Continue supporting current Rust versions in the default configuration
2. On targets where std is missing, compile as no_std using the feature detection leo added in autocfg (Requires Rust 1.36)
3. There's no feature flag for this, because it's not needed (negative feature flags being awkward), and because it could be considered to be a breaking change
4. Constructor `new` etc is not available without std — one has to use the others like `with_hasher` or `default`. The `indexmap!` macros depend on these as well, and require std too.

This approach, in particular (3) is novel, and thus it is out for wider requests for comments.

A no-std indexmap has been requested by petgraph.

Fixes #94 
Closes #95 
Closes #92

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,15 @@ matrix:
     - rust: nightly
       env:
        - FEATURES='test_low_transition_point'
-    - name: "no_std"
-      rust: stable
+    - rust: 1.36.0
+      env: TARGET=thumbv6m-none-eabi
+      before_script:
+        - rustup target add $TARGET
+        - set -ex
+      script:
+        - cargo build -vv --target=$TARGET
+        - cargo build -v -p test-nostd --target=$TARGET
+    - rust: stable
       env: TARGET=thumbv6m-none-eabi
       before_script:
         - rustup target add $TARGET
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,15 @@ matrix:
     - rust: nightly
       env:
        - FEATURES='test_low_transition_point'
+    - name: "no_std"
+      rust: stable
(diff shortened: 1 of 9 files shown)


bluss:
I also looked at autocfg to figure out how to test this.

I have a worry that other "no_std" using crates are going to be disappointed in this (3) scheme, when they can't test this with crate features as we normally do.

I still lean towards this being the morally right way to handle no_std; we build without it because it is not available.

I'm looking for a better no-std hasher than ahash (since it depends on a lot of stuff).

cuviper:
The main hazard I see in leaving this implicit is that users could end up depending on `std`-required API (like `IndexMap::new`) without realizing it. I guess it's up to them to try a true `no_std` target if they really need that. Plus this kind of failure can also happen with unified features -- if something else in your dependency tree enables a feature, you can also use it without declaring so.

But I am in the camp that losing `std` with `default-features = false` is a breaking change, so this auto-detection is really the only way around that.

> I'm looking for a simpler no-std hasher than ahash (since it depends on a lot of stuff).

Maybe [`twox-hash`](https://crates.io/crates/twox-hash)? Or since it's just for testing, you could add a private copy of `fnv` / `fxhash`.

bluss:
Thanks for implementing @leo60228 and thanks @cuviper for guiding the implementation.

I'll aim to release this in Indexmap 1.3 since it's a new feature

PR #112, opened 2020-01-12
Upgrade rand and quickcheck

kjeremy:


bluss:
Thanks. This PR is on hold, because it breaks the build on Rust 1.30, which needs further action - increment minor version for example.

bluss:
Ah, it was a minor mistake to use "closes" in the other PR (That closes this PR with red status, but just having github detect it was merged elsewhere would close it in purple status).

PR #113, opened 2020-01-14
Fixed clippy warnings

Razican:
This should improve the efficiency a bit, and the code readability in the `resolve()` function, with one less tab.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -124,12 +124,12 @@ impl Pos {
     fn none() -> Self { Pos { index: !0 } }
 
     #[inline]
-    fn is_none(&self) -> bool { self.index == !0 }
+    fn is_none(self) -> bool { self.index == !0 }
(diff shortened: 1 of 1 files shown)


cuviper:
Can you run some before/after benchmarks? Especially if you could try both 32-bit and 64-bit targets. I expect it's fine for 64-bit targets where `Pos` is the same size as the reference (pointer) we're avoiding, but it technically makes the argument larger on 32-bit targets. That may not matter though -- these should all be inlined, so the parameter mode becomes irrelevant.

Razican:
Actually, after doing some benchmarks, it's a bit (but statistically relevant) slower in x86-64 targets. Didn't check a 32-bit target.

bluss:
Thanks, we don't need to investigate this further then, I think, and can skip this PR

PR #115, opened 2020-01-31
Adds entry_index to map.rs

Thermatix:
I wanted a way to access the index but the only way to do that is  by getting the entry and then the index from that which is mutable (and causes issues) or from `get_full` which returns more than what I needed so I added `entry_index`

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1032,7 +1032,7 @@ where
     }
 
     /// Return item index
-    pub fn get_index<Q: ?Sized>(&self, key: &Q) -> Option<usize>
+    pub fn entry_index<Q: ?Sized>(&self, key: &Q) -> Option<usize>
     where
         Q: Hash + Equivalent<K>,
     {
--- a/src/map.rs
+++ b/src/map.rs
@@ -1037,7 +1037,6 @@ where
         Q: Hash + Equivalent<K>,
     {
         if let Some((_, found)) = self.find(key) {
-            let entry = &self.core.entries[found];
             Some(found)
         } else {
             None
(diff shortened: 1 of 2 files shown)


src/map.rs
```diff
@@ -1031,6 +1031,19 @@ where
         }
     }
 
+    /// Return item index
+    pub fn entry_index<Q: ?Sized>(&self, key: &Q) -> Option<usize>
+    where
+        Q: Hash + Equivalent<K>,
+    {
+        if let Some((_, found)) = self.find(key) {
+            let entry = &self.core.entries[found];
```
cuviper: Please remove the unused variable.

Thermatix:
@cuviper, I'm sorry I haven't gotten back to you on this, I was sick over the last week and just genuinely forgot, I'll get on this as soon as I'm able!

Thermatix:
Ok, Made the requested changes!

bluss:
Wait - this method seems underspecified and surprising? Both the PR and the doc comment needs to explain in detail what it does.

cuviper:
Well, at least it's not published yet if you want to change/revert it.

Yes, the main advantage I see is to avoid the indexing operation in `get_full`, which seems reasonable to me. The author originally called it `get_index`, which might be a better name, but that method already exists for _using_ an index to get the key-value.

bluss:
I am sorry for my tone there, that's certainly not pitch perfect, and extremely inappropriate for a "maintainer" that is as AWOL as I am.

Name doesn't need to be perfect, but it's clear that the current name is confusing, and it would be best to find a better doc comment or name before release - but it's not a requirement either.

PR #118, opened 2020-03-27
feat: add OccupiedEntry.get_entry_mut, VacantEntry.insert_entry

jhgg:
Adds 2 utility functions ontop of the Entry API.

- `OccupiedEntry.get_entry_mut` - for when you want a reference to the key, and a mutable reference to the value.
- `VacantEntry.insert_entry` - for when you want to insert into the map, and return a reference to the key, and a mutable reference to the value.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -693,6 +693,10 @@ impl<'a, K, V> OccupiedEntry<'a, K, V> {
     pub fn get_mut(&mut self) -> &mut V {
         &mut self.map.entries[self.index].value
     }
+    pub fn get_entry_mut(&mut self) -> (&K, &mut V) {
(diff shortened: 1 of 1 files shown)


cuviper:
Has anything like this been proposed for the `std` maps? Ideally, we would keep a compatible API, and these methods don't seem particular to `IndexMap`.

jhgg:
No :( I just needed these for a project I'm working on - and felt like it'd be cool to upstream it here. I don't have the time to get these into std tho.

glaebhoerl:
A similar API was recently added to hashbrown: https://github.com/rust-lang/hashbrown/pull/145

PR #126, opened 2020-06-05
Unify the entry and insert code

### LINKED ISSUE #108: Improve .entry() performance
In some of the benchmarks, if I replace `.insert()` with code based on entry, the benchmarks run slower. If we could get these to performance parity, then we can delete the `.insert()` code, since it is duplicated by the functionality in `.entry()`.

To test this, you can update insert to be like this (note `insert_full` uses `entry` right now):

```rust
    pub fn insert(&mut self, key: K, value: V) -> Option<V> {
        self.insert_full(key, value).1
    }
```

The following is the benchmark difference I see when using `entry` in `insert`.

The hashmap benchmarks can be ignored - ordermap is the interesting part. (The hashmap benchmarks are controls.)

Change from before to after. 
```
 name                                   63 ns/iter  62 ns/iter  diff ns/iter  diff % 
 insert_hashmap_100_000                 2,488,863   2,504,113         15,250   0.61% 
 insert_hashmap_10_000                  214,543     221,693            7,150   3.33% 
 insert_hashmap_150                     3,210       3,339                129   4.02% 
 insert_hashmap_int_bigvalue_10_000     262,701     264,518            1,817   0.69% 
 insert_hashmap_str_10_000              236,377     247,805           11,428   4.83% 
 insert_hashmap_string_10_000           1,117,104   1,148,170         31,066   2.78% 
 insert_hashmap_string_oneshot_10_000   1,078,875   1,085,733          6,858   0.64% 
 insert_orderedmap_100_000              2,727,119   3,874,446      1,147,327  42.07% 
 insert_orderedmap_10_000               267,361     363,741           96,380  36.05% 
 insert_orderedmap_150                  3,956       5,664              1,708  43.17% 
 insert_orderedmap_int_bigvalue_10_000  359,582     413,168           53,586  14.90% 
 insert_orderedmap_str_10_000           305,463     312,993            7,530   2.47% 
 insert_orderedmap_string_10_000        1,066,189   1,091,526         25,337   2.38%
```

mwillsey:
This will close #108 

This doesn't so much improve the entry code rather than abstract away `entry` and `insert` into a `ProbeAction`. This forces monomorphization to give us the insert fast path that doesn't require the additional indirection that entry does.

It also implements `insert_full` on that fast path (`insert` is now just a call to `insert_full`), so that should be much faster as well, since it was formerly implemented in terms of `entry`.

Benchmarks
```
 name                                   control ns/iter  variable ns/iter  diff ns/iter  diff %  speedup 
 insert_orderedmap_100_000              2,688,143        2,616,739              -71,404  -2.66%   x 1.03 
 insert_orderedmap_10_000               275,198          272,283                 -2,915  -1.06%   x 1.01 
 insert_orderedmap_150                  4,116            3,862                     -254  -6.17%   x 1.07 
 insert_orderedmap_int_bigvalue_10_000  409,681          395,290                -14,391  -3.51%   x 1.04 
 insert_orderedmap_str_10_000           298,722          294,373                 -4,349  -1.46%   x 1.01 
 insert_orderedmap_string_10_000        2,027,784        1,948,038              -79,746  -3.93%   x 1.04 
```

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1700,7 +1700,7 @@ trait ProbeAction<'a, Sz: Size, K, V>: Sized {
     fn hit(self, entry: OccupiedEntry<'a, K, V>) -> Self::Output;
     // handle an empty spot in the map
     fn empty(self, entry: VacantEntry<'a, K, V>) -> Self::Output;
-    // robin hood: handle a that you should steal because it's better for you
+    // robin hood: handle a spot that you should steal because it's better for you
     fn steal(self, entry: VacantEntry<'a, K, V>) -> Self::Output;
 }
 
--- a/src/map.rs
+++ b/src/map.rs
@@ -351,12 +351,6 @@ fn probe_distance(mask: usize, hash: HashValue, current: usize) -> usize {
     current.wrapping_sub(desired_pos(mask, hash)) & mask
 }
 
-enum Inserted<V> {
-    Done,
-    Swapped { prev_value: V },
-    RobinHood { probe: usize, old_pos: Pos },
-}
-
 impl<K, V, S> fmt::Debug for IndexMap<K, V, S>
 where
     K: fmt::Debug + Hash + Eq,
@@ -840,13 +834,13 @@ where
     K: Hash + Eq,
     S: BuildHasher,
 {
-    // FIXME: reduce duplication (compare with insert)
-    fn entry_phase_1<Sz>(&mut self, key: K) -> Entry<K, V>
+    fn probe_action<'a, Sz, A>(&'a mut self, key: K, action: A) -> A::Output
     where
         Sz: Size,
+        A: ProbeAction<'a, Sz, K, V>,
     {
         let hash = hash_elem_using(&self.hash_builder, &key);
-        self.core.entry_phase_1::<Sz>(hash, key)
+        self.core.probe_action::<Sz, A>(hash, key, action)
     }
 
     /// Remove all key-value pairs in the map, while preserving its capacity.
@@ -865,24 +859,12 @@ where
         }
     }
 
-    // First phase: Look for the preferred location for key.
-    //
-    // We will know if `key` is already in the map, before we need to insert it.
-    // When we insert they key, it might be that we need to continue displacing
-    // entries (robin hood hashing), in which case Inserted::RobinHood is returned
-    fn insert_phase_1<Sz>(&mut self, key: K, value: V) -> Inserted<V>
-    where
-        Sz: Size,
-    {
-        let hash = hash_elem_using(&self.hash_builder, &key);
-        self.core.insert_phase_1::<Sz>(hash, key, value)
-    }
-
     fn reserve_one(&mut self) {
         if self.len() == self.capacity() {
             dispatch_32_vs_64!(self.double_capacity());
         }
     }
+
     fn double_capacity<Sz>(&mut self)
     where
         Sz: Size,
@@ -904,26 +886,7 @@ where
     /// See also [`entry`](#method.entry) if you you want to insert *or* modify
     /// or if you need to get the index of the corresponding key-value pair.
     pub fn insert(&mut self, key: K, value: V) -> Option<V> {
-        self.reserve_one();
-        if self.size_class_is_64bit() {
-            match self.insert_phase_1::<u64>(key, value) {
-                Inserted::Swapped { prev_value } => Some(prev_value),
-                Inserted::Done => None,
-                Inserted::RobinHood { probe, old_pos } => {
-                    self.core.insert_phase_2::<u64>(probe, old_pos);
-                    None
-                }
-            }
-        } else {
-            match self.insert_phase_1::<u32>(key, value) {
-                Inserted::Swapped { prev_value } => Some(prev_value),
-                Inserted::Done => None,
-                Inserted::RobinHood { probe, old_pos } => {
-                    self.core.insert_phase_2::<u32>(probe, old_pos);
-                    None
-                }
-            }
-        }
+        self.insert_full(key, value).1
     }
 
     /// Insert a key-value pair in the map, and get their index.
@@ -940,16 +903,8 @@ where
     /// See also [`entry`](#method.entry) if you you want to insert *or* modify
     /// or if you need to get the index of the corresponding key-value pair.
     pub fn insert_full(&mut self, key: K, value: V) -> (usize, Option<V>) {
-        let entry = self.entry(key);
-        let index = entry.index();
-
-        match entry {
-            Entry::Occupied(mut entry) => (index, Some(entry.insert(value))),
-            Entry::Vacant(entry) => {
-                entry.insert(value);
-                (index, None)
-            }
-        }
+        self.reserve_one();
+        dispatch_32_vs_64!(self.probe_action::<_>(key, InsertValue(value)))
     }
 
     /// Get the given key’s corresponding entry in the map for insertion and/or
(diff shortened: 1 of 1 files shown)


src/map.rs
```diff
@@ -1785,6 +1694,59 @@ impl<K, V> OrderMapCore<K, V> {
...
+        *pos = Pos::with_hash::<Sz>(index, entry.hash);
+        entry.map.entries.push(Bucket {
+            hash: entry.hash,
+            key: entry.key,
+            value: self.0,
+        });
+        (index, None)
+    }
+    fn steal(self, entry: VacantEntry<'a, K, V>) -> Self::Output {
+        let index = entry.map.entries.len();
```
bluss, line 1726: this could be just `entry.insert_impl::<Sz>(self.0)` right? Something like that to reduce duplication

bluss:
That's awesome

src/map.rs
```diff
@@ -1785,6 +1694,59 @@ impl<K, V> OrderMapCore<K, V> {
     }
 }
 
+trait ProbeAction<'a, Sz: Size, K, V>: Sized {
+    type Output;
+    fn hit(self, entry: OccupiedEntry<'a, K, V>) -> Self::Output;
```
bluss, line 1699: If possible, document each method with a one-line summary of what it represents

src/map.rs
```diff
@@ -1785,6 +1694,59 @@ impl<K, V> OrderMapCore<K, V> {
...
+
+struct InsertValue<V>(V);
+
+impl<'a, Sz: Size, K, V> ProbeAction<'a, Sz, K, V> for InsertValue<V> {
+    type Output = (usize, Option<V>);
+    fn hit(self, entry: OccupiedEntry<'a, K, V>) -> Self::Output {
+        let old = replace(&mut entry.map.entries[entry.index].value, self.0);
+        (entry.index, Some(old))
+    }
+    fn empty(self, entry: VacantEntry<'a, K, V>) -> Self::Output {
```
bluss, line 1714: A style nitpick is that all the methods need an empty line separating them from the item before them

src/map.rs
```diff
@@ -840,13 +834,13 @@ where
     K: Hash + Eq,
     S: BuildHasher,
 {
-    // FIXME: reduce duplication (compare with insert)
-    fn entry_phase_1<Sz>(&mut self, key: K) -> Entry<K, V>
+    fn probe_action<'a, Sz, A>(&'a mut self, key: K, action: A) -> A::Output
```
bluss, line 837: Nice refactoring. I think this method needs a clearer name. We use "probe" in every loop when reading/writing to the map, so it is not specific enough. Even "insert phase 1" is an ok name - and `insert_phase_2` still exists, so it needs to be renamed in tandem, if we entirely remove anything named "phase 1".

src/map.rs
```diff
@@ -1785,6 +1694,59 @@ impl<K, V> OrderMapCore<K, V> {
     }
 }
 
+trait ProbeAction<'a, Sz: Size, K, V>: Sized {
+    type Output;
+    fn hit(self, entry: OccupiedEntry<'a, K, V>) -> Self::Output;
+    fn empty(self, entry: VacantEntry<'a, K, V>) -> Self::Output;
+    fn steal(self, entry: VacantEntry<'a, K, V>) -> Self::Output {
+        self.empty(entry)
+    }
```
bluss, line 1703: For code readability we can prefer to not have this default impl for `steal`.

src/map.rs
mwillsey: Ah good idea. Do you mean `empty` could use that? I was afraid that entering the probe loop another time would be costly (technically, you don't have to in the empty case), but it seems to be fine.

mwillsey: Nvm, turns out doing it in `empty` isn't great, it basically remove the fast path that the old insert code had, making steal and empty the same.

bluss: Are we then maybe identifying the main thing that slows down entry compared with the old insert? Maybe if we updated entry's VacantEntry to skip that loop, for the non-steal case, that insert and entry would match each other in performance?

bluss:
Could you update the PR description to use one of the formulations that means that #108 is closed when this is merged? :) https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

src/map.rs
```diff
@@ -1785,6 +1694,60 @@ impl<K, V> OrderMapCore<K, V> {
     }
 }
 
+trait ProbeAction<'a, Sz: Size, K, V>: Sized {
+    type Output;
+    // handle an occupied spot in the map
+    fn hit(self, entry: OccupiedEntry<'a, K, V>) -> Self::Output;
+    // handle an empty spot in the map
+    fn empty(self, entry: VacantEntry<'a, K, V>) -> Self::Output;
+    // robin hood: handle a that you should steal because it's better for you
```
cuviper, line 1703:
```suggestion
    // robin hood: handle a spot that you should steal because it's better for you
```

bluss:
Thanks. I'll merge this when I have time to make some post-merge edits

cuviper:
@bluss can I help with whatever is blocking this?

bluss:
I'll add it as comments. I thought I might as well merge and edit myself, sometimes I prefer that to the back-and-forth, just don't want reviews to be too perfectionist

src/map.rs
```diff
@@ -1785,6 +1694,60 @@ impl<K, V> OrderMapCore<K, V> {
...
+    fn empty(self, entry: VacantEntry<'a, K, V>) -> Self::Output {
+        let pos = &mut entry.map.indices[entry.probe];
+        let index = entry.map.entries.len();
+        *pos = Pos::with_hash::<Sz>(index, entry.hash);
+        entry.map.entries.push(Bucket {
+            hash: entry.hash,
+            key: entry.key,
+            value: self.0,
+        });
+        (index, None)
```
bluss, line 1726: This method body seems "loose" like this - could it not just call a method on the entry? I general, it's hard to follow the insert logic when it's spread out.

(This comment is rather perfectionist, and not something I wanted to make a requirement to fix before merging).

src/map.rs
```diff
@@ -1785,6 +1694,60 @@ impl<K, V> OrderMapCore<K, V> {
...
+}
+
+struct MakeEntry;
+
+impl<'a, Sz: Size, K: 'a, V: 'a> ProbeAction<'a, Sz, K, V> for MakeEntry {
+    type Output = Entry<'a, K, V>;
+    fn hit(self, entry: OccupiedEntry<'a, K, V>) -> Self::Output {
+        Entry::Occupied(entry)
+    }
+    fn empty(self, entry: VacantEntry<'a, K, V>) -> Self::Output {
```
bluss, line 1743: nitpick missing empty line between the methods

cuviper:
Yeah, I get that, and sometimes I'll take advantage of maintainer pushes to PRs for that same reason.

I'm pressing because I'm building up changes toward [using indexmap more in rustc](https://github.com/rust-lang/rust/issues/60608). :slightly_smiling_face:

src/map.rs
cuviper: I looked at this -- you'd basically want `VacantEntry::insert_impl` without the probe loop (since we know its empty). But this needs to consume the entry's key, which makes it harder to reuse that code, so I think we'd end up writing this code separately either way. i.e. It doesn't really matter whether we do it here or there.

(And frankly, a lot of this will disappear if I land the hashbrown changes...)

cuviper:
I fixed the spacing nit, and I think we can wait on the other point.

Thanks @mwillsey!

PR #128, opened 2020-06-08
Adding `reverse()` to `IndexMap` & `IndexSet`

### LINKED ISSUE #127: Adding `.reverse()` method to `IndexMap` & `IndexSet`
Looking at the impl of `OrderMapCore`, `.sort_by()` saves the hash index, sorts the underlying `entries` vector, and then restores the hash index. Would this mean that a `.reverse()` method could be added to `OrderedMapCore` (and thus `IndexMap`/`IndexSet`)?

linclelinkpart5:
Closes #127.

Here's my work so far, just basing it off of what's done in `sort_by()`. Happy to include the more efficient re-indexing as you suggested, @cuviper! Just need to learn how.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1760,9 +1760,22 @@ impl<K, V> OrderMapCore<K, V> {
     }
 
     fn reverse(&mut self) {
-        let side_index = self.save_hash_index();
         self.entries.reverse();
-        self.restore_hash_index(side_index);
+
+        // No need to save hash indices, can easily calculate what they should
+        // be, given that this is an in-place reversal.
+        dispatch_32_vs_64!(self => apply_new_index(&mut self.indices, self.entries.len()));
+
+        fn apply_new_index<Sz>(indices: &mut [Pos], len: usize)
+        where
+            Sz: Size,
+        {
+            for pos in indices {
+                if let Some((i, _)) = pos.resolve::<Sz>() {
+                    pos.set_pos::<Sz>(len - i);
+                }
+            }
+        }
     }
 
     fn save_hash_index(&mut self) -> Vec<usize> {
--- a/src/map.rs
+++ b/src/map.rs
@@ -1253,6 +1253,13 @@ where
         self.into_iter()
     }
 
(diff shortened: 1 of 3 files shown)


src/map.rs
```diff
@@ -1253,6 +1253,13 @@ where
         self.into_iter()
     }
 
+    /// Reverses the order of the map’s key-value pairs in place.
+    ///
+    /// Computes in **O(n)** time and **O(1)** space.
```
cuviper, line 1258: O(1) space isn't true when we're calling `save_hash_index` because that's O(n) -- but the optimization we discussed will avoid the need for that.

The idea is to mimic `restore_hash_map`'s `apply_new_index`, but changing this:

```rust
    pos.set_pos::<Sz>(new_index[i]);
```

We know what the new reversed index will be without a saved table, just:

```rust
    pos.set_pos::<Sz>(len - i);
```

bluss:
Please update the PR description to close the feature request with a keyword https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword thanks

src/map.rs
cuviper: Pardon me -- that should be `len - i - 1`!

linclelinkpart5: I think I'm doing something wrong: using both `self.entries.len()` and `self.indices.len()` as `len` gives me out of bounds errors.

linclelinkpart5: Aha, that was the cause, curse you off-by-one errors! Using `self.entries.len() - i - 1` works wonderfully.

cuviper: Yeah, `self.entries.len()` is correct, but I was off by one for the new index.

bluss:
Special thanks for using quickcheck

linclelinkpart5:
Thanks all! And no problem @bluss, it was fun to work with. It inspired me to look into it for testing my current project, so it's a win-win!

PR #130, opened 2020-06-09
Use a consistently seeded Rng for benchmark stability

cuviper:
I was seeing a lot of variance in benchmark results, and I noticed that
we were using `SmallRng::from_entropy()`. Now we use a fixed seed value,
so we test the same input data every time.


cuviper:
This only affects the benchmarks, and CI passed, so I'll go ahead and merge.

bluss:
Thanks, good catch.

PR #131, opened 2020-06-09
Switch to hashbrown's RawTable internally

cuviper:
This switches `IndexMapCore` from its bespoke hash table to `hashbrown::raw::RawTable<usize>`, storing just the index into our ordered `entries: Vec<Bucket<K, V>>`. We lose the badge of having no `unsafe` code, but the overall implementation is much simpler, relying on the battle-hardened `hashbrown` for the tricky parts. I have also confirmed that the testsuite passes under `cargo miri`.

I'll post the benchmark comparison in a followup comment, but broadly speaking it appears faster for insertion and lookup, and slower for removal. I think that's a reasonable trade-off for this crate.

As a bonus, I also implemented a proper `reserve` and added `shrink_to_fit`, since they are now pretty straightforward.

cc @Amanieu -- thanks for `hashbrown` and for exposing `RawTable` to make this possible!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,11 @@ autocfg = "1"
 serde = { version = "1.0", optional = true, default-features = false }
 rayon = { version = "1.0", optional = true }
 
+[dependencies.hashbrown]
+version = "0.7"
+default-features = false
+features = ["inline-more", "raw"]
+
 [dev-dependencies]
 itertools = "0.8"
 rand = {version = "0.7", features = ["small_rng"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ serde = { version = "1.0", optional = true, default-features = false }
 rayon = { version = "1.0", optional = true }
 
 [dependencies.hashbrown]
-version = "0.7"
+version = "0.8"
 default-features = false
 features = ["inline-more", "raw"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ default-features = false
 features = ["inline-more", "raw"]
 
 [dev-dependencies]
-itertools = "0.8"
+itertools = "0.9"
 rand = {version = "0.7", features = ["small_rng"] }
 quickcheck = { version = "0.9", default-features = false }
 fnv = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ rayon = { version = "1.0", optional = true }
 [dependencies.hashbrown]
 version = "0.8"
 default-features = false
-features = ["inline-more", "raw"]
+features = ["raw"]
 
 [dev-dependencies]
 itertools = "0.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "indexmap"
-version = "1.4.0"
+version = "1.5.0"
 authors = [
 "bluss",
 "Josh Stone <cuviper@gmail.com>"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ serde = { version = "1.0", optional = true, default-features = false }
 rayon = { version = "1.0", optional = true }
 
 [dependencies.hashbrown]
-version = "0.8"
+version = "0.8.1"
 default-features = false
 features = ["raw"]
 
--- a/src/map.rs
+++ b/src/map.rs
@@ -2,8 +2,6 @@
 //! pairs is independent of the hash values of the keys.
 
 #[cfg(not(has_std))]
-use alloc::boxed::Box;
-#[cfg(not(has_std))]
 use std::vec::Vec;
 
 pub use mutable_keys::MutableKeys;
@@ -29,12 +27,6 @@ use util::third;
 use {Bucket, Entries, HashValue};
 
 pub use map_core::{Entry, OccupiedEntry, VacantEntry};
-
-fn hash_elem_using<B: BuildHasher, K: ?Sized + Hash>(build: &B, k: &K) -> HashValue {
-    let mut h = build.build_hasher();
-    k.hash(&mut h);
-    HashValue(h.finish() as usize)
-}
 
 /// A hash table where the iteration order of the key-value pairs is independent
 /// of the hash values of the keys.
@@ -110,14 +102,17 @@ where
 impl<K, V, S> Entries for IndexMap<K, V, S> {
     type Entry = Bucket<K, V>;
 
+    #[inline]
     fn into_entries(self) -> Vec<Self::Entry> {
         self.core.into_entries()
     }
 
+    #[inline]
     fn as_entries(&self) -> &[Self::Entry] {
         self.core.as_entries()
     }
 
+    #[inline]
     fn as_entries_mut(&mut self) -> &mut [Self::Entry] {
         self.core.as_entries_mut()
     }
@@ -137,18 +132,21 @@ where
     S: BuildHasher,
 {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.debug_map().entries(self.iter()).finish()?;
-        if cfg!(feature = "test_debug") {
-            writeln!(f)?;
-            self.core.debug_entries(f)?;
+        if cfg!(not(feature = "test_debug")) {
+            f.debug_map().entries(self.iter()).finish()
+        } else {
+            // Let the inner `IndexMapCore` print all of its details
+            f.debug_struct("IndexMap")
+                .field("core", &self.core)
+                .finish()
         }
-        Ok(())
     }
 }
 
 #[cfg(has_std)]
 impl<K, V> IndexMap<K, V> {
     /// Create a new map. (Does not allocate.)
+    #[inline]
     pub fn new() -> Self {
         Self::with_capacity(0)
     }
@@ -157,6 +155,7 @@ impl<K, V> IndexMap<K, V> {
     /// allocate if `n` is zero.)
     ///
     /// Computes in **O(n)** time.
+    #[inline]
     pub fn with_capacity(n: usize) -> Self {
         Self::with_capacity_and_hasher(n, <_>::default())
     }
@@ -167,6 +166,7 @@ impl<K, V, S> IndexMap<K, V, S> {
     /// allocate if `n` is zero.)
     ///
     /// Computes in **O(n)** time.
+    #[inline]
     pub fn with_capacity_and_hasher(n: usize, hash_builder: S) -> Self
     where
         S: BuildHasher,
@@ -187,6 +187,7 @@ impl<K, V, S> IndexMap<K, V, S> {
     /// Return the number of key-value pairs in the map.
     ///
     /// Computes in **O(1)** time.
+    #[inline]
     pub fn len(&self) -> usize {
         self.core.len()
     }
@@ -194,6 +195,7 @@ impl<K, V, S> IndexMap<K, V, S> {
     /// Returns true if the map contains no elements.
     ///
     /// Computes in **O(1)** time.
+    #[inline]
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
@@ -234,11 +236,22 @@ where
 
     /// Reserve capacity for `additional` more key-value pairs.
     ///
-    /// FIXME Not implemented fully yet.
+    /// Computes in **O(n)** time.
     pub fn reserve(&mut self, additional: usize) {
-        if additional > 0 {
-            self.core.reserve_one();
-        }
+        self.core.reserve(additional);
+    }
+
+    /// Shrink the capacity of the map as much as possible.
+    ///
+    /// Computes in **O(n)** time.
+    pub fn shrink_to_fit(&mut self) {
+        self.core.shrink_to_fit();
+    }
+
+    fn hash<Q: ?Sized + Hash>(&self, key: &Q) -> HashValue {
+        let mut h = self.hash_builder.build_hasher();
+        key.hash(&mut h);
+        HashValue(h.finish() as usize)
     }
 
     /// Insert a key-value pair in the map.
@@ -272,7 +285,7 @@ where
     /// See also [`entry`](#method.entry) if you you want to insert *or* modify
     /// or if you need to get the index of the corresponding key-value pair.
     pub fn insert_full(&mut self, key: K, value: V) -> (usize, Option<V>) {
-        let hash = hash_elem_using(&self.hash_builder, &key);
+        let hash = self.hash(&key);
         self.core.insert_full(hash, key, value)
     }
 
@@ -281,7 +294,7 @@ where
     ///
     /// Computes in **O(1)** time (amortized average).
     pub fn entry(&mut self, key: K) -> Entry<K, V> {
-        let hash = hash_elem_using(&self.hash_builder, &key);
+        let hash = self.hash(&key);
         self.core.entry(hash, key)
     }
 
@@ -328,7 +341,7 @@ where
     where
         Q: Hash + Equivalent<K>,
     {
-        self.find(key).is_some()
+        self.get_index_of(key).is_some()
     }
 
     /// Return a reference to the value stored for `key`, if it is present,
@@ -339,7 +352,12 @@ where
     where
         Q: Hash + Equivalent<K>,
     {
-        self.get_full(key).map(third)
+        if let Some(i) = self.get_index_of(key) {
+            let entry = &self.as_entries()[i];
+            Some(&entry.value)
+        } else {
+            None
+        }
     }
 
     /// Return references to the key-value pair stored for `key`,
@@ -363,9 +381,9 @@ where
     where
         Q: Hash + Equivalent<K>,
     {
-        if let Some(found) = self.get_index_of(key) {
-            let entry = &self.as_entries()[found];
-            Some((found, &entry.key, &entry.value))
+        if let Some(i) = self.get_index_of(key) {
+            let entry = &self.as_entries()[i];
+            Some((i, &entry.key, &entry.value))
         } else {
             None
         }
@@ -376,10 +394,11 @@ where
     where
         Q: Hash + Equivalent<K>,
     {
-        if let Some((_, found)) = self.find(key) {
-            Some(found)
-        } else {
+        if self.is_empty() {
             None
+        } else {
+            let hash = self.hash(key);
+            self.core.get_index_of(hash, key)
         }
     }
 
@@ -387,14 +406,24 @@ where
     where
         Q: Hash + Equivalent<K>,
     {
-        self.get_full_mut(key).map(third)
+        if let Some(i) = self.get_index_of(key) {
+            let entry = &mut self.as_entries_mut()[i];
+            Some(&mut entry.value)
+        } else {
+            None
+        }
     }
 
     pub fn get_full_mut<Q: ?Sized>(&mut self, key: &Q) -> Option<(usize, &K, &mut V)>
     where
         Q: Hash + Equivalent<K>,
     {
-        self.get_full_mut2_impl(key).map(|(i, k, v)| (i, &*k, v))
+        if let Some(i) = self.get_index_of(key) {
+            let entry = &mut self.as_entries_mut()[i];
+            Some((i, &entry.key, &mut entry.value))
+        } else {
+            None
+        }
     }
 
     pub(crate) fn get_full_mut2_impl<Q: ?Sized>(
@@ -404,25 +433,12 @@ where
     where
         Q: Hash + Equivalent<K>,
     {
-        if let Some((_, found)) = self.find(key) {
-            let entry = &mut self.as_entries_mut()[found];
-            Some((found, &mut entry.key, &mut entry.value))
+        if let Some(i) = self.get_index_of(key) {
+            let entry = &mut self.as_entries_mut()[i];
+            Some((i, &mut entry.key, &mut entry.value))
         } else {
             None
         }
-    }
-
-    /// Return probe (indices) and position (entries)
-    fn find<Q: ?Sized>(&self, key: &Q) -> Option<(usize, usize)>
-    where
-        Q: Hash + Equivalent<K>,
-    {
-        if self.is_empty() {
-            return None;
-        }
-        let h = hash_elem_using(&self.hash_builder, key);
-        self.core
-            .find_using(h, move |entry| Q::equivalent(key, &entry.key))
     }
 
     /// Remove the key-value pair equivalent to `key` and return
@@ -504,12 +520,11 @@ where
     where
         Q: Hash + Equivalent<K>,
     {
-        let (probe, found) = match self.find(key) {
-            None => return None,
-            Some(t) => t,
-        };
-        let (k, v) = self.core.swap_remove_found(probe, found);
-        Some((found, k, v))
+        if self.is_empty() {
+            return None;
+        }
+        let hash = self.hash(key);
+        self.core.swap_remove_full(hash, key)
     }
 
     /// Remove the key-value pair equivalent to `key` and return
@@ -562,19 +577,18 @@ where
     where
         Q: Hash + Equivalent<K>,
     {
-        let (probe, found) = match self.find(key) {
-            None => return None,
-            Some(t) => t,
-        };
-        let (k, v) = self.core.shift_remove_found(probe, found);
-        Some((found, k, v))
+        if self.is_empty() {
+            return None;
+        }
+        let hash = self.hash(key);
+        self.core.shift_remove_full(hash, key)
     }
 
     /// Remove the last key-value pair
     ///
     /// Computes in **O(1)** time (average).
     pub fn pop(&mut self) -> Option<(K, V)> {
-        self.core.pop_impl()
+        self.core.pop()
     }
 
     /// Scan through each key-value pair in the map and keep those where the
@@ -588,7 +602,7 @@ where
     where
         F: FnMut(&K, &mut V) -> bool,
     {
-        self.retain_mut(move |k, v| keep(k, v));
+        self.core.retain_in_order(move |k, v| keep(k, v));
     }
 
     pub(crate) fn retain_mut<F>(&mut self, keep: F)
@@ -605,7 +619,9 @@ where
     where
         K: Ord,
     {
-        self.core.sort_by(key_cmp)
+        self.with_entries(|entries| {
+            entries.sort_by(|a, b| Ord::cmp(&a.key, &b.key));
+        });
     }
 
     /// Sort the map’s key-value pairs in place using the comparison
@@ -616,11 +632,13 @@ where
     ///
     /// Computes in **O(n log n + c)** time and **O(n)** space where *n* is
     /// the length of the map and *c* the capacity. The sort is stable.
-    pub fn sort_by<F>(&mut self, compare: F)
+    pub fn sort_by<F>(&mut self, mut cmp: F)
     where
         F: FnMut(&K, &V, &K, &V) -> Ordering,
     {
-        self.core.sort_by(compare)
+        self.with_entries(|entries| {
+            entries.sort_by(|a, b| cmp(&a.key, &a.value, &b.key, &b.value));
+        });
     }
 
     /// Sort the key-value pairs of the map and return a by value iterator of
@@ -632,7 +650,7 @@ where
         F: FnMut(&K, &V, &K, &V) -> Ordering,
     {
         let mut entries = self.into_entries();
-        entries.sort_by(move |a, b| cmp(&a.key, &a.value, &b.key, &b.value));
+        entries.sort_by(|a, b| cmp(&a.key, &a.value, &b.key, &b.value));
         IntoIter {
             iter: entries.into_iter(),
         }
@@ -652,13 +670,6 @@ where
             iter: self.core.drain(range),
         }
     }
-}
-
-fn key_cmp<K, V>(k1: &K, _v1: &V, k2: &K, _v2: &V) -> Ordering
-where
-    K: Ord,
-{
-    Ord::cmp(k1, k2)
 }
 
 impl<K, V, S> IndexMap<K, V, S> {
@@ -690,11 +701,7 @@ impl<K, V, S> IndexMap<K, V, S> {
     ///
     /// Computes in **O(1)** time (average).
     pub fn swap_remove_index(&mut self, index: usize) -> Option<(K, V)> {
-        let (probe, found) = match self.as_entries().get(index) {
-            Some(e) => self.core.find_existing_entry(e),
-            None => return None,
-        };
-        Some(self.core.swap_remove_found(probe, found))
+        self.core.swap_remove_index(index)
     }
 
     /// Remove the key-value pair by index
@@ -707,11 +714,7 @@ impl<K, V, S> IndexMap<K, V, S> {
     ///
     /// Computes in **O(n)** time (average).
     pub fn shift_remove_index(&mut self, index: usize) -> Option<(K, V)> {
-        let (probe, found) = match self.as_entries().get(index) {
-            Some(e) => self.core.find_existing_entry(e),
-            None => return None,
-        };
-        Some(self.core.shift_remove_found(probe, found))
+        self.core.shift_remove_index(index)
     }
 }
 
@@ -1020,11 +1023,7 @@ where
 
     /// ***Panics*** if `key` is not present in the map.
     fn index(&self, key: &'a Q) -> &V {
-        if let Some(v) = self.get(key) {
-            v
-        } else {
-            panic!("IndexMap: key not found")
-        }
+        self.get(key).expect("IndexMap: key not found")
     }
 }
 
@@ -1040,11 +1039,7 @@ where
 {
     /// ***Panics*** if `key` is not present in the map.
     fn index_mut(&mut self, key: &'a Q) -> &mut V {
-        if let Some(v) = self.get_mut(key) {
-            v
-        } else {
-            panic!("IndexMap: key not found")
-        }
+        self.get_mut(key).expect("IndexMap: key not found")
     }
 }
 
@@ -1082,9 +1077,21 @@ where
     /// equivalents of a key occur more than once, the last corresponding value
     /// prevails.
     fn extend<I: IntoIterator<Item = (K, V)>>(&mut self, iterable: I) {
-        for (k, v) in iterable {
+        // (Note: this is a copy of `std`/`hashbrown`'s reservation logic.)
+        // Keys may be already present or show multiple times in the iterator.
+        // Reserve the entire hint lower bound if the map is empty.
+        // Otherwise reserve half the hint (rounded up), so the map
+        // will only resize twice in the worst case.
+        let iter = iterable.into_iter();
+        let reserve = if self.is_empty() {
+            iter.size_hint().0
+        } else {
+            (iter.size_hint().0 + 1) / 2
+        };
+        self.reserve(reserve);
+        iter.for_each(move |(k, v)| {
             self.insert(k, v);
-        }
+        });
     }
 }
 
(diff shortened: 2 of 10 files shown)


cuviper:
Note that this does raise the MSRV to 1.32, but that's well more than a year old. Perhaps we should release 1.4.1 for other recent changes, then bump to 1.5.0 for this change. We can also upgrade to 2018 edition, but I didn't want to pollute this PR with that churn.

<details> 
  <summary><code>cargo benchcmp master hashbrown</code></summary>
<pre>
 name                                         master ns/iter  hashbrown ns/iter  diff ns/iter   diff %  speedup
 entry_hashmap_150                            2,533           2,476                       -57   -2.25%   x 1.02
 entry_indexmap_150                           3,809           2,976                      -833  -21.87%   x 1.28
 few_retain_hashmap_100_000                   1,016,146       1,005,607               -10,539   -1.04%   x 1.01
 few_retain_indexmap_100_000                  999,829         490,861                -508,968  -50.91%   x 2.04
 grow_fnv_hashmap_100_000                     1,995,356       1,757,197              -238,159  -11.94%   x 1.14
 grow_fnv_indexmap_100_000                    2,959,801       2,424,241              -535,560  -18.09%   x 1.22
 half_retain_hashmap_100_000                  973,803         1,005,521                31,718    3.26%   x 0.97
 half_retain_indexmap_100_000                 1,131,965       900,842                -231,123  -20.42%   x 1.26
 hashmap_merge_shuffle                        494,258         467,592                 -26,666   -5.40%   x 1.06
 hashmap_merge_simple                         370,419         384,379                  13,960    3.77%   x 0.96
 indexmap_clone_for_sort_s                    370,587         377,810                   7,223    1.95%   x 0.98
 indexmap_clone_for_sort_u32                  4,987           19,638                   14,651  293.78%   x 0.25
 indexmap_merge_shuffle                       432,270         356,746                 -75,524  -17.47%   x 1.21
 indexmap_merge_simple                        361,191         269,597                 -91,594  -25.36%   x 1.34
 indexmap_simple_sort_s                       1,890,796       1,953,812                63,016    3.33%   x 0.97
 indexmap_simple_sort_u32                     646,615         631,954                 -14,661   -2.27%   x 1.02
 indexmap_sort_s                              1,597,762       1,591,000                -6,762   -0.42%   x 1.00
 indexmap_sort_u32                            520,058         526,665                   6,607    1.27%   x 0.99
 insert_hashmap_100_000                       2,179,030       2,364,903               185,873    8.53%   x 0.92
 insert_hashmap_10_000                        195,234         209,327                  14,093    7.22%   x 0.93
 insert_hashmap_150                           2,923           3,116                       193    6.60%   x 0.94
 insert_hashmap_int_bigvalue_10_000           229,529         232,528                   2,999    1.31%   x 0.99
 insert_hashmap_str_10_000                    194,549         196,252                   1,703    0.88%   x 0.99
 insert_hashmap_string_10_000                 970,037         944,748                 -25,289   -2.61%   x 1.03
 insert_hashmap_string_10_000                 878,069         923,558                  45,489    5.18%   x 0.95
 insert_hashmap_string_oneshot_10_000         929,238         941,393                  12,155    1.31%   x 0.99
 insert_indexmap_100_000                      2,125,932       2,246,446               120,514    5.67%   x 0.95
 insert_indexmap_10_000                       213,993         189,754                 -24,239  -11.33%   x 1.13
 insert_indexmap_150                          3,210           2,926                      -284   -8.85%   x 1.10
 insert_indexmap_int_bigvalue_10_000          282,849         262,640                 -20,209   -7.14%   x 1.08
 insert_indexmap_str_10_000                   226,982         207,124                 -19,858   -8.75%   x 1.10
 insert_indexmap_string_10_000                881,052         879,212                  -1,840   -0.21%   x 1.00
 insert_indexmap_string_10_000                936,610         882,226                 -54,384   -5.81%   x 1.06
 iter_black_box_hashmap_10_000                18,282          18,481                      199    1.09%   x 0.99
 iter_black_box_indexmap_10_000               3,028           3,007                       -21   -0.69%   x 1.01
 iter_sum_hashmap_10_000                      10,546          10,419                     -127   -1.20%   x 1.01
 iter_sum_indexmap_10_000                     2,045           2,016                       -29   -1.42%   x 1.01
 lookup_hashmap_100_000_inorder_multi         68,768          69,260                      492    0.72%   x 0.99
 lookup_hashmap_100_000_multi                 72,178          70,700                   -1,478   -2.05%   x 1.02
 lookup_hashmap_100_000_single                15              16                            1    6.67%   x 0.94
 lookup_hashmap_10_000_exist                  68,488          68,528                       40    0.06%   x 1.00
 lookup_hashmap_10_000_exist_string           122,444         123,332                     888    0.73%   x 0.99
 lookup_hashmap_10_000_exist_string_oneshot   101,518         101,856                     338    0.33%   x 1.00
 lookup_hashmap_10_000_noexist                66,051          66,800                      749    1.13%   x 0.99
 lookup_indexmap_100_000_inorder_multi        82,405          87,341                    4,936    5.99%   x 0.94
 lookup_indexmap_100_000_multi                124,984         121,523                  -3,461   -2.77%   x 1.03
 lookup_indexmap_100_000_single               25              24                           -1   -4.00%   x 1.04
 lookup_indexmap_10_000_exist                 119,710         89,257                  -30,453  -25.44%   x 1.34
 lookup_indexmap_10_000_exist_string          176,353         165,409                 -10,944   -6.21%   x 1.07
 lookup_indexmap_10_000_exist_string_oneshot  160,672         131,624                 -29,048  -18.08%   x 1.22
 lookup_indexmap_10_000_noexist               120,897         74,732                  -46,165  -38.19%   x 1.62
 many_retain_hashmap_100_000                  374,578         381,946                   7,368    1.97%   x 0.98
 many_retain_indexmap_100_000                 891,856         1,165,316               273,460   30.66%   x 0.77
 new_hashmap                                  5               5                             0    0.00%   x 1.00
 new_indexmap                                 4               6                             2   50.00%   x 0.67
 pop_indexmap_100_000                         1,079,201       1,611,619               532,418   49.33%   x 0.67
 shift_remove_indexmap_100_000_few            14,464,051      15,427,596              963,545    6.66%   x 0.94
 shift_remove_indexmap_2_000_full             2,188,620       2,642,177               453,557   20.72%   x 0.83
 swap_remove_indexmap_100_000                 3,587,036       4,523,900               936,864   26.12%   x 0.79
 with_capacity_10e5_hashmap                   562             556                          -6   -1.07%   x 1.01
 with_capacity_10e5_indexmap                  1,511           215                      -1,296  -85.77%   x 7.03
</pre>
</details>

cuviper:
For future work, I've also been playing with parameterizing the index type, since `rustc` likes to use newtype-`u32` indexes. We can extend to something like this without a breaking change:

```rust
pub struct IndexMap<K, V, S = RandomState, Idx = usize> { ... }
pub struct IndexSet<T, S = RandomState, Idx = usize> { ... }
```

Then internally we'd switch to `RawTable<Idx>`.

src/map.rs
```diff
@@ -548,11 +564,13 @@ where
...
     /// the length of the map and *c* the capacity. The sort is stable.
-    pub fn sort_by<F>(&mut self, compare: F)
+    pub fn sort_by<F>(&mut self, mut cmp: F)
     where
         F: FnMut(&K, &V, &K, &V) -> Ordering,
     {
-        self.core.sort_by(compare)
+        self.with_entries(|entries| {
+            entries.sort_by(|a, b| cmp(&a.key, &a.value, &b.key, &b.value));
+        });
```
bluss, line 573: style-wise I'd put `move` on both these closures so that they capture the closure by value; just something we've seen to optimize better in some cases

src/map.rs
```diff
@@ -952,11 +955,7 @@ where
 
     /// ***Panics*** if `key` is not present in the map.
     fn index(&self, key: &'a Q) -> &V {
-        if let Some(v) = self.get(key) {
-            v
-        } else {
-            panic!("IndexMap: key not found")
-        }
+        self.get(key).expect("IndexMap: key not found")
```
bluss, line 958: Style-wise I'm not sure I agree that Option::expect is a good way for a library to report a panic, I prefer the panic as it was

src/map_core.rs
```diff
@@ -1,406 +1,101 @@
+#![allow(unsafe_code)]
```
bluss, line 1: We need to add a comment to the top-level deny we have for this - to say that we have some sections where it is allowed in the core of the crate.

bluss:
Really cool work.

So we lose the 32-bit index optimization, and still have some performance improvements, does that mean that there is more to gain?

We are giving up being implemented in safe Rust and can only do so if we show performance improvements that are above the noise. Some of the lookup benchmark cases do that, and the insert cases barely do it. The improved lookup benches are really encouraging.

I'll read RawTable and then come back to reviewing. Are you going to start experimenting with this version of indexmap in rustc, before it gets merged?

cuviper:
@bluss where do you stand on this? Just waiting for time to review it?

bluss:
This weekend should have some time where I can review

Cargo.toml
```diff
@@ -35,8 +35,13 @@ autocfg = "1"
 serde = { version = "1.0", optional = true, default-features = false }
 rayon = { version = "1.0", optional = true }
 
+[dependencies.hashbrown]
+version = "0.8"
+default-features = false
+features = ["inline-more", "raw"]
```
bluss, line 41: rustc doesn't want to use `inline-more`, so it seems like we shouldn't

src/map_core.rs
```diff
@@ -409,585 +104,291 @@ impl<K, V> Entries for IndexMapCore<K, V> {
...
-            let mut probe = desired_pos(self.mask, entry_hash);
-            probe_loop!(probe < self.indices.len(), {
-                if self.indices[probe].is_none() {
-                    // empty bucket, insert here
-                    self.indices[probe] = Pos::with_hash::<SzNew>(i, entry_hash);
-                    return;
-                }
-            });
+    /// Append a key-value pair, *without* checking whether it already exists.
+    fn push(&mut self, hash: HashValue, key: K, value: V) -> usize {
```
bluss, line 180: Doc could explain the return value

src/map.rs
```diff
@@ -360,25 +378,36 @@ where
     where
         Q: Hash + Equivalent<K>,
     {
-        if let Some((_, found)) = self.find(key) {
-            Some(found)
-        } else {
+        if self.is_empty() {
```
bluss, line 381: Why check for empty here?

src/map_core.rs
```diff
@@ -409,585 +104,291 @@ impl<K, V> Entries for IndexMapCore<K, V> {
...
-                    if i > found {
-                        // shift the index
-                        pos.set_pos::<Sz>(i - 1);
+        // use a heuristic between a full sweep vs. a `find()` for every shifted item.
+        let raw_capacity = self.indices.buckets();
+        let shifted_entries = &self.entries[index..];
+        if shifted_entries.len() > raw_capacity / 2 {
+            // shift all indices greater than `index`
+            unsafe {
+                for raw_bucket in self.indices.iter() {
```
bluss, line 284: Ok, so the reason `RawTable::iter` is `unsafe` is because it has no lifetime parameter, it's a proxy for a bunch of raw pointers. However as long as we have the right borrowing behavior, there are no other traps or requirements on calling `next`, IIUC.

src/map_core.rs
```diff
@@ -409,585 +104,291 @@ impl<K, V> Entries for IndexMapCore<K, V> {
...
-        self.reserve_one();
-        dispatch_32_vs_64!(self.insert_phase_1::<_>(hash, key, MakeEntry))
-    }
-
-    /// Return probe (indices) and position (entries)
-    pub(crate) fn find_using<F>(&self, hash: HashValue, key_eq: F) -> Option<(usize, usize)>
+        match self.find_equivalent(hash, &key) {
+            Some(raw_bucket) => Entry::Occupied(OccupiedEntry {
+                map: self,
+                raw_bucket,
```
bluss, line 220: I don't think this is very easy to analyze but here we have `map: &'a mut IndexMapCore<K, V>` and a `RawBucket` and the raw bucket is used as a mutable pointer into something that is part of the map, on the surface level this is a borrowing error, and I don't know the rule that would allow this, off hand.

src/map_core.rs
bluss: `raw_bucket` here shadows the parameter `raw_bucket` in a way that obscures the logic, so I think it should use a different name. Same in the else branch.

bluss:
There isn't much `unsafe` code, that's nice, but the questions of simultaneous reference and raw pointer validity stump me at the moment, so it's not trivial code.

src/map_core.rs
bluss: On a closer look, `fn shift_remove_bucket(&mut self, raw_bucket: RawBucket)` also runs into the same question, of course.

Cargo.toml
cuviper: I'll benchmark the difference, at least.

I'm not sure we can actually "pollute" `std`'s feature choices, because that's a separate cargo run from where `indexmap` is used in `rustc` crates. Right now we have a different version anyway, but I'm not actually sure what will happen when those align -- will cargo try to rebuild all of `std` when we enable the "raw" feature?

src/map_core.rs
cuviper: Will do -- it's just the new index, a slight optimization to not have to figure it from the length again. At least this is internal, otherwise it might be questionable that this is different than `Vec::push`...

src/map.rs
cuviper: It's a small perf hack to avoid computing the hash when there's nothing to compare. The former `fn find` did the same.

src/map_core.rs
cuviper: > However as long as we have the right borrowing behavior, there are no other traps or requirements on calling `next`, IIUC.

Yes, where the "right" behavior avoids anything that would invalidate pointers, namely reallocating.

> `raw_bucket` here shadows the parameter

OK, I'll rename these.

src/map_core.rs
cuviper: Superficially, the API forces us to do this -- [`erase_no_drop`](https://docs.rs/hashbrown/0.8.0/hashbrown/raw/struct.RawTable.html#method.erase_no_drop) takes `&mut self` and `&Bucket<T>`.

I believe this is sound because the buckets are not directly part of the `RawTable`, but rather part of a separate allocation that the table also holds by raw pointer. It's fine for that to alias the raw pointer in the `Bucket`, and we don't form any references through those pointers. Miri is happy with it, both in `hashbrown`'s CI and when I ran `indexmap`'s testsuite manually under miri.

Cargo.toml
cuviper: Here are my results regarding "inline-more" -- it's not an obvious win after all!

<details> 
  <summary><code>cargo benchcmp default inline-more</code></summary>
<pre>
 name                                         default ns/iter  inline-more ns/iter  diff ns/iter   diff %  speedup
 entry_indexmap_150                           3,379            2,932                        -447  -13.23%   x 1.15
 few_retain_indexmap_100_000                  464,831          548,747                    83,916   18.05%   x 0.85
 grow_fnv_indexmap_100_000                    2,493,403        2,417,127                 -76,276   -3.06%   x 1.03
 half_retain_indexmap_100_000                 1,189,542        1,174,703                 -14,839   -1.25%   x 1.01
 indexmap_clone_for_sort_s                    380,182          412,785                    32,603    8.58%   x 0.92
 indexmap_clone_for_sort_u32                  20,595           23,040                      2,445   11.87%   x 0.89
 indexmap_merge_shuffle                       371,277          367,470                    -3,807   -1.03%   x 1.01
 indexmap_merge_simple                        291,215          291,980                       765    0.26%   x 1.00
 indexmap_simple_sort_s                       2,112,893        1,990,345                -122,548   -5.80%   x 1.06
 indexmap_simple_sort_u32                     665,741          642,336                   -23,405   -3.52%   x 1.04
 indexmap_sort_s                              1,619,554        1,633,687                  14,133    0.87%   x 0.99
 indexmap_sort_u32                            539,317          538,438                      -879   -0.16%   x 1.00
 insert_indexmap_100_000                      2,186,996        2,194,670                   7,674    0.35%   x 1.00
 insert_indexmap_10_000                       196,985          190,736                    -6,249   -3.17%   x 1.03
 insert_indexmap_150                          3,048            2,917                        -131   -4.30%   x 1.04
 insert_indexmap_int_bigvalue_10_000          271,094          272,833                     1,739    0.64%   x 0.99
 insert_indexmap_str_10_000                   221,377          206,219                   -15,158   -6.85%   x 1.07
 insert_indexmap_string_10_000                931,370          857,500                   -73,870   -7.93%   x 1.09
 insert_indexmap_string_10_000                937,381          845,038                   -92,343   -9.85%   x 1.11
 iter_black_box_indexmap_10_000               3,038            3,130                          92    3.03%   x 0.97
 iter_sum_indexmap_10_000                     2,040            2,092                          52    2.55%   x 0.98
 lookup_indexmap_100_000_inorder_multi        86,897           90,024                      3,127    3.60%   x 0.97
 lookup_indexmap_100_000_multi                116,602          121,846                     5,244    4.50%   x 0.96
 lookup_indexmap_100_000_single               23               25                              2    8.70%   x 0.92
 lookup_indexmap_10_000_exist                 88,555           88,473                        -82   -0.09%   x 1.00
 lookup_indexmap_10_000_exist_string          171,322          165,636                    -5,686   -3.32%   x 1.03
 lookup_indexmap_10_000_exist_string_oneshot  143,021          138,956                    -4,065   -2.84%   x 1.03
 lookup_indexmap_10_000_noexist               70,532           71,557                      1,025    1.45%   x 0.99
 many_retain_indexmap_100_000                 1,259,902        1,354,056                  94,154    7.47%   x 0.93
 new_indexmap                                 7                6                              -1  -14.29%   x 1.17
 pop_indexmap_100_000                         1,661,312        1,701,130                  39,818    2.40%   x 0.98
 shift_remove_indexmap_100_000_few            16,781,653       16,870,955                 89,302    0.53%   x 0.99
 shift_remove_indexmap_2_000_full             2,511,175        2,729,205                 218,030    8.68%   x 0.92
 swap_remove_indexmap_100_000                 4,583,354        4,640,754                  57,400    1.25%   x 0.99
 with_capacity_10e5_indexmap                  203              201                            -2   -0.99%   x 1.01
</pre>
</details>

cuviper: I just tested with `std` updated to the same hashbrown 0.8.0. It does build two copies of `hashbrown`, but this seems OK. For `std`->`hashbrown` it has "rustc-dep-of-std" enabled, with associated unstable features. For `rustc`->`indexmap`->`hashbrown`, it has "raw" and (for now) "inline-more". AFAICS this works fine.

bluss: Great, let's drop it, Alex Crichton delivered some good rationale for avoiding this feature anyway, no reason not to follow this decision (see hashbrown PR history).

cuviper: OK, I'll remove "inline-more". For reference, I suppose you're referring to https://github.com/rust-lang/hashbrown/pull/119

src/map_core.rs
bluss: This is just that there is a lot happening in unsafe coding guidelines. Stacked borrows model, rules around raw pointers and pointer provenance. In some cases, a lot more things are forbidden than I thought were with raw pointers. Miri doesn't check raw pointers very well, I think.

I'm sorry, I just don't have a model that I can follow here. That the API forces it is unfortunately not an argument, the other arguments are appreciated 😄  With this code, I feel unmoored and don't know how to think about it - every argument about validity needs to start with the implementation details of RawTable (I can read those), and what do the `&mut self` and `&self` signatures on the methods then even mean anymore?

I'll go incoherent I fear. On a slice it's important that we have `as_mut_ptr(&mut self)` because we cast `&mut [T] as *mut [T] as *mut T`. On a `Vec<T>` is it important we modify through `as_mut_ptr`? Why is there a story that modification through [`Vec::as_ptr -> *const T` is UB](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_ptr)? `RawTable::data_start(&self)` is the same kind of accessor to me.

cuviper: I'm not sure exactly what the rules are about mutating through a `*const T` -- `NonNull` is a notable exception, but maybe this is just an example where `std` can break the rules since it's coupled with the compiler. That is, the result of such UB may be "works as expected" (for now), but if the compiler ever changes (and `NonNull` with it) then that UB could turn nasty to outsiders trying this.

@RalfJung added that warning on slice `as_ptr` (https://github.com/rust-lang/rust/pull/60443) and later copied to `Vec::as_ptr` (https://github.com/rust-lang/rust/pull/61114) -- maybe he can clarify the actual rules?

The [reference](https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html) says, "all data reached through a shared reference or data owned by an immutable binding is immutable, unless that data is contained within an `UnsafeCell<U>`." In that case, I think hashbrown may have a real problem -- e.g. `HashMap::get_mut(&mut self, ...)` calls `RawTable::find(&self, ...)` to get a bucket, and then forms a `&mut V` from that. Does that origin through `&self` render the bucket pointer immutable?

CAD97: `Vec::as_ptr` and `Vec::as_mut_ptr` have the exact same implementation. They both just return the raw pointer held by `RawVec` (notably, via `RawVec::ptr(&self) -> *mut T`), and thus return pointers with the same provenance. The only reason you aren't allowed to mutate through `Vec::as_ptr` is that the docs say so. And I agree that they should; `Vec` is a safe abstraction, so should obey the typical safe rules. `RawVec` is an unsafe abstraction, so obeys the raw pointer rules.

The reference's "data reached through a shared reference" only refers to data behind shared references, and not data behind raw pointers. The important part here is that `RawVec`/`RawTable` (they play the same role here) are raw pointer abstractions.

To be clear, I don't think this understanding of the rules is _explicitly_ condoned by Rust the language. However, a _lot_ of code, both inside and outside std, is written with this interpretation of the rules.

cuviper: > `Vec::as_ptr` and `Vec::as_mut_ptr` have the exact same implementation. They both just return the raw pointer held by `RawVec` (notably, via `RawVec::ptr(&self) -> *mut T`), and thus return pointers with the same provenance.

Peeking at the implementation isn't really informative if that's another "std is special" situation. It tells you that there are no dragons lurking _yet_, but they could hatch in the future. That's why the doc statement is important.

RalfJung: > added that warning on slice as_ptr (rust-lang/rust#60443) and later copied to Vec::as_ptr (rust-lang/rust#61114) -- maybe he can clarify the actual rules?

There is no official aliasing model yet, and as such no official rules. But I can explain what [Stacked Borrows](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md) says.

There, the rules are that *which initial "raw borrow" you use* matters, but later the type of the raw ptr does not matter any more. So if `x` is a reference, if you do `x as *mut _` (which is sugar for `&raw mut *x`) then the resulting pointer may be used for mutation. If you do `x as *const _` (sugar for `&raw const *x`), the resulting pointer may *not* be used for mutation.

Also see https://github.com/rust-lang/rust/issues/56604 and in particular [these notes](https://hackmd.io/wpBesJ8PSRG0dX4LHcUZUQ).

RalfJung: > The only reason you aren't allowed to mutate through Vec::as_ptr is that the docs say so.

In particular, as usual with library UB, the `Vec` implementation is allowed to change in the future in a way that actually causes language UB when you mutate through `as_ptr`.

bluss: @RalfJung thanks for commenting. You address the initial raw borrow, but in this case the life of the raw pointer, just comes from the allocator I think? (Just like in Vec, technically)?  In this example, there is never a case where we consider the data pointers to be derived from the `&RawTable` or `&mut RawTable` references at all.

This is the critical piece I think I'm missing - when do we talk about about raw pointers being based on a &/&mut borrow at all?

It seems like in this case, we get a raw pointer from the allocator and we copy the pointer along (we can use an accessor on &RawTable to copy the raw pointer), and it is considered valid to write through this raw pointer all the way through?

RalfJung: > This is the critical piece I think I'm missing - when do we talk about about raw pointers being based on a &/&mut borrow at all?

When they are cast from a *reference* (`&T`/`&mut T`).

The initial pointer obtained from the allocator is mutable. It doesnt matter how the raw pointer gets copied around, as long as it does not get turned into a reference and then back into a raw pointer (at that point, the cast rules mentioned above apply).

bluss: Thanks, previously I've been wondering a lot about what cuviper quotes, the "all data reached through a shared reference or data owned by an immutable binding is immutable, unless that data is contained within an UnsafeCell<U>."

But it is clarified here that in the accessor for example `RawTable::data_end(&self) -> *mut T` the raw pointer is not considered to be based on the `&RawTable` reference.

In this quote's formulation, it has been unclear what "based on" means and I think that's just due to that it hasn't really been formally specified at all, I know it's an ongoing process. <3

RalfJung: Yeah... "based on" is basically the same as [in LLVM](https://llvm.org/docs/LangRef.html#pointer-aliasing-rules).

So, when you compute `p2` from `p1` by doing offsets, then `p2` is "based on" `p1`. But when you *load* `p2` from memory by dereferencing `p1`, that does not induce any kind of "based on" relationship.

bluss: I don't think the notion of me being confused is fair, when this is a whole bundle of unspecified and not yet formalized rules. From my perspective we're hanging on tight and trying to adapt unsafe code to what the future rules will be, based on ongoing discussion.

RalfJung: Sorry, I didn't mean to imply anything negative here.

Pointer "based on" being somehow affected by loads from pointers-to-pointers is a common misconception (but one I did not at all foresee).

bluss: For pedagogical consideration - for the rest of the community, I mean - I'll mention that I think there is a strong will to have `&mut self/&self` be meaningful, and we are sometimes thinking of this on a Rust conceptual level, not just loading a pointer through a pointer, but want to assign a meaning to accessing through `&mut self` vs `&self`.

RalfJung: Fair. OTOH, I think there are also many people that expect this propagation of meaning to stop at raw pointers, and I think this is indeed required to make raw pointers even useful.

bluss:
Thanks all for your input on the discussion, and helping me find some solid footing again with the raw pointer code. Thanks cuviper for working on this - I'll work on helping you wrap this up as soon as you want.

There are two discomforting things left; I think it's something that's rather general to come up in this situation

- We handle raw buckets (raw pointer wrappers) that are passed in structs and between function scopes

   - Usually in Rust, we use safe abstractions, maybe just a simple lifetime to help us with the book keeping when leaving local reasoning. Lifetimes often help us convert the problem into just using local reasoning instead of global reasoning.
      I'm not saying it's worth the trouble to wrap this in further abstractions here, to make it safer, but it's the things we try to do if it's possible.

- We have internal functions that don't follow the usual safe/unsafe distinction

  - I believe it is best practice, that for example `fn swap_remove_bucket(&mut self, raw_bucket: RawBucket) -> (usize, K, V)`,
    should be an unsafe method. The reason is that we must trust the caller to pass a correct raw bucket.
    I'll push a commit with this change. History is mutable, so feel free to toss it out if you don't like it.

  - We end up in a gray area here, since struct fields can not be made unsafe to assign (known issue), so we don't follow this completely stringently.

bluss:
This might be the first time we raise the MSRV in 1.x, but with a good reason.

It looks like we might break serde_json, which builds CI with Rust 1.31? cc @dtolnay, when we raise ours to Rust 1.32 here.

Other top rdeps - toml, http, petgraph, they look fine from MSRV standpoint.

cuviper:
> * We handle raw buckets (raw pointer wrappers) that are passed in structs and between function scopes
> 
>     * Usually in Rust, we use safe abstractions, maybe just a simple lifetime to help us with the book keeping when leaving local reasoning. Lifetimes often help us convert the problem into just using local reasoning instead of global reasoning.
>       I'm not saying it's worth the trouble to wrap this in further abstractions here, to make it safer, but it's the things we try to do if it's possible.

I'll play a little with abstracting this. It would be nice if we could contain `unsafe` even more tightly.

> * We have internal functions that don't follow the usual safe/unsafe distinction
> 
>     * I believe it is best practice, that for example `fn swap_remove_bucket(&mut self, raw_bucket: RawBucket) -> (usize, K, V)`,
>       should be an unsafe method. The reason is that we must trust the caller to pass a correct raw bucket.
>       I'll push a commit with this change.
>
>     * We end up in a gray area here, since struct fields can not be made unsafe to assign (known issue), so we don't follow this completely stringently.

Yeah, this is kind of like the [tootsie-pop model](https://smallcultfollowing.com/babysteps/blog/2016/05/27/the-tootsie-pop-model-for-unsafe-code/), just like `Vec`'s length and capacity are innocent-looking fields with important safety implications.

> This might be the first time we raise the MSRV in 1.x, but with a good reason.

If we entertain the idea of indexmap 2.0, are there other changes you would want?

> It looks like we might break serde_json, which builds CI with Rust 1.31?

I am reluctant to push MSRV changes on others, but at least that is only for their non-default `preserve_order` feature. Still, they have CI to test that on 1.31, so they'd have to at least hold that back in CI to keep working. Such downgrades are possible, but annoying to deal with.

cuviper:
> I'll push a commit with this change.

That looks fine, thanks.

bluss:
I don't think it's wrong to carefully raise the MSRV, it's the plan we have documented and promised since 1.0. However we can discuss it with those that depend on us.

cuviper:
I tried an abstraction to wrap the raw bucket with an appropriate lifetime, by bundling to a map reference. We really only need that in `OccupiedEntry`, and I found that this structure basically _is_ that wrapper already, so trying to add a separate layer just adds confusing indirection.

I settled instead on just encapsulating this in its own module. I moved `map_core` to `map::core`, then added a further `map::core::raw` for anything that has to deal with raw buckets, which is the only `unsafe` code.

bluss:
I agree

bluss:
I can't really think of any breaking changes that are on the wishlist.

I definitely don't want to pile on work here, but a 2.0 version could include the parameterization by index type. Then we can update the `Default` impl to be fully generic without having to fear type inference regressions (and the other `new*` methods too?).

The "experimental things - Equivalent trait works well, the MutableKeys trait I don't know, so those things seem like they could stay for 2.0.

Which method to make the default `.remove()` - I still think the current decision is fine, no change from my end for 2.0.
(Good way to document this would be to say that "remove is the fastest way to remove an element, and if order should be preserved, ...")

src/map/core/raw.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// Remove an entry by shifting all entries that follow it
+    ///
+    /// Safety: The caller must pass a live `raw_bucket`.
+    #[allow(unused_unsafe)]
+    unsafe fn shift_remove_bucket(&mut self, raw_bucket: RawBucket) -> (usize, K, V) {
+        // use Vec::remove, but then we need to update the indices that point
+        // to all of the other entries that have to move
+        let index = unsafe {
+            self.indices.erase_no_drop(&raw_bucket);
+            raw_bucket.read()
```
jonhoo, line 104: I don't think this matters are the moment, but you should probably do the `read` before you erase the bucket, since _technically_ the map can do whatever it wants to the bucket once you've erased it. I ran into this when trying to use [`griddle`] as a drop-in replacement from `hashbrown`, since it _does_ occasionally invalidate the bucket's memory when it is erased.

src/map/core/raw.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// Remove an entry by swapping it with the last
+    ///
+    /// Safety: The caller must pass a live `raw_bucket`.
+    #[allow(unused_unsafe)]
+    unsafe fn swap_remove_bucket(&mut self, raw_bucket: RawBucket) -> (usize, K, V) {
+        // use swap_remove, but then we need to update the index that points
+        // to the other entry that has to move
+        let index = unsafe {
+            self.indices.erase_no_drop(&raw_bucket);
+            raw_bucket.read()
```
jonhoo, line 165: Same here — the read should happen before the erase.

src/map/core/raw.rs
cuviper: Hmm, `hashbrown` uses this pattern itself a lot, e.g. in `OccupiedEntry`:

```rust
    pub fn remove_entry(self) -> (K, V) {
        unsafe {
            self.table.table.erase_no_drop(&self.elem);
            self.elem.read()
        }
    }
```

@Amanieu, can you clarify if this is something we can rely on? We should probably update the documentation of `erase_no_drop` either way.

> since it _does_ occasionally invalidate the bucket's memory when it is erased.

I can't see how that would happen, unless you allow other mutating operations before you actually read it.

Not that it's a big deal to reorder this, just a little more awkward.

jonhoo: So, in `griddle` the reason this can happen is that `erase_no_drop` may erase the last non-empty bucket from a table that we were migrating items away from. If that is the case, then that is the best time for griddle to de-allocate it. It _could_ realize this later if we strictly need the contract for `erase_no_drop` to be this way, but it seems nicer if `erase_no_drop` isn't _required_ to continue to leave the bucket available after it is called.

Amanieu: `erase_no_drop` only marks the bucket as deleted in the control bytes. It doesn't touch the actual data and it is fine to read from it. In fact hashbrown relies on this when dropping values in-place since `drop_in_place` may panic but we still want to ensure the bucket is marked as deleted if that happens.

cuviper:
> I definitely don't want to pile on work here, but a 2.0 version could include the parameterization by index type. Then we can update the `Default` impl to be fully generic without having to fear type inference regressions (and the other `new*` methods too?).

Well, I do want to tackle that next, and I've even prototyped it. I was hoping a default `Idx = usize` would let it be free of breaking changes, but I can see how `Default` type inference may be a problem. If we embrace 2.0, that would give more freedom in how we add it -- e.g. should it be `IndexMap<K, V, S, Idx>` or `IndexMap<K, V, Idx, S>`?

We could also choose a higher MSRV for more guarantees, like 1.36 to always have `alloc` -- although we'd still need to probe/feature `std` for `RandomState`.

src/map/core/raw.rs
jonhoo: I guess my argument here is that while you _can_ rely on it for `hashbrown`, it'd be nice if that wasn't a part of `erase_no_drop`'s _contract_, since it means that anyone who wants to mimick must also uphold that contract. Or alternatively, it'd be great if `indexmap` did not rely on this aspect of `erase_no_drop` (though admittedly for selfish reasons). It's of course fine for `hashbrown` to continue to depend on this internally, since it's not going to use anyone else's `raw` any time soon :p

In griddle, I got around `drop_in_place` by instead doing a `read` + `erase_no_drop` + `drop`. It's not _quite_ as efficient, but it does mean that `erase_no_drop` is free to reclaim memory immediately should it so wish.

jonhoo: For what it's worth, I also agree with @cuviper that if this is indeed something that `erase_no_drop` guarantees, then it should be documented that that is the case, especially since the name does not suggest that that guarantee exists. If `erase_no_drop` is updated that way, then I'll also update `griddle` to match, even if it adds some reclamation overhead.

jonhoo: I filed this as https://github.com/rust-lang/hashbrown/issues/168. Depending on the resolution to that, this code should either stay the way it is (if it becomes a part of `erase_no_drop`'s guarantees), or the order of the operations should be reversed (if it does not).

cuviper: > `erase_no_drop` may erase the last non-empty bucket from a table that we were migrating items away from. If that is the case, then that is the best time for griddle to de-allocate it.

This feels a little contradictory. I guess you can narrowly interpret "no drop" to refer to the item only, but dropping the allocation is still a relatively heavy operation for what the method name implies.

jonhoo: Yeah, I suppose that's true. I almost wonder if `erase_no_drop` should be entirely replace with something like `erase_and_take`? Or maybe `erase_and_take` + `erase_and_drop`.

jonhoo: Now that https://github.com/rust-lang/hashbrown/pull/171 has landed, I think this should probably be merged as-is (to avoid holding it up), and then we should move to `erase`/`remove` once `hashbrown 0.8.1` is out.

jonhoo:
Should the discussion of what else to land for a potential 2.0 happen separately from this PR? Might be good to merge this and then do the follow-up discussion in an issue?

src/map/core.rs
```diff
@@ -0,0 +1,409 @@
...
+        self.indices.len()
+    }
+
+    #[inline]
+    pub(crate) fn capacity(&self) -> usize {
+        cmp::min(self.indices.capacity(), self.entries.capacity())
+    }
+
+    pub(crate) fn clear(&mut self) {
+        self.indices.clear_no_drop();
```
jonhoo, line 131: Is there a particular reason you use `clear_no_drop` over `clear` here? The latter already optimizes for the case when the type doesn't need dropping: https://docs.rs/hashbrown/0.8.0/src/hashbrown/raw/mod.rs.html#612

cuviper: I suppose that's an overly-eager optimization on my part. Would it help your griddle port to use plain `clear`?

jonhoo: It would — `clear_no_drop` is tricky in griddle since it's not clear how you clear the iterator of "leftover" elements without also dropping the backing storage for that iterator.

cuviper: Done.

cuviper:
I opened #135 for 2.0 discussion.

bluss:
I'd vote to merge this and go for indexmap 1.5 with this

jonhoo:
Not sure if my vote counts, but I agree :p

cuviper:
I'm fine with that.

For new APIs in the release notes, we have `reverse` from #128, and this PR's `shrink_to_fit` and improved `reserve`. Maybe `clone_from` in #125 is worth mentioning, just to know that you can now clone into an existing allocation.

cuviper:
I've added the version bump and some release notes to this PR.

Amanieu:
`insert_no_grow` is now available in hashbrown 0.8.1.

cuviper:
Great! I've updated to use the new hashbrown methods.

cuviper:
Let's ship it!

PR #132, opened 2020-06-11
Implement direct usize indexing

cuviper:
```rust
where
    IndexMap<K, V, S>: IndexMut<usize, Output = V>,
    IndexSet<T, S>: Index<usize, Output = T>,
```

This allows `map[i]` and `set[i]` indexing to access values directly,
panicking if the index is out of bounds, similar to slices.

On maps, this somewhat overlaps with `Index<&Q> + IndexMut<&Q>` where
`Q: Equivalent<K>`. The reference makes this indexing unambiguous, but
it could be confusing to users if the key type is also an integer.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1023,6 +1023,30 @@ where
     }
 }
 
+impl<K, V, S> Index<usize> for IndexMap<K, V, S> {
+    type Output = V;
+
+    /// ***Panics*** if `index` is out of bounds.
+    fn index(&self, index: usize) -> &V {
+        self.get_index(index)
+            .expect("IndexMap: index out of bounds")
+            .1
+    }
+}
+
+/// Mutable indexing allows changing / updating indexed values
+/// that are already present.
+///
+/// You can **not** insert new values with index syntax, use `.insert()`.
+impl<K, V, S> IndexMut<usize> for IndexMap<K, V, S> {
+    /// ***Panics*** if `index` is out of bounds.
+    fn index_mut(&mut self, index: usize) -> &mut V {
+        self.get_index_mut(index)
+            .expect("IndexMap: index out of bounds")
+            .1
+    }
+}
+
 impl<K, V, S> FromIterator<(K, V)> for IndexMap<K, V, S>
 where
     K: Hash + Eq,
--- a/src/map.rs
+++ b/src/map.rs
@@ -993,6 +993,28 @@ impl<K, V, S> IntoIterator for IndexMap<K, V, S> {
     }
 }
 
+/// Access `IndexMap` values corresponding to a key.
+///
+/// # Examples
+///
+/// ```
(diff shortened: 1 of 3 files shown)


cuviper:
In my experiments in rustc, replacing `HashMap,Vec` combinations with `indexmap`, I've found it annoying to replace direct `vec[i]` indexing with `get_index(i).unwrap().1`. I didn't find anywhere that we've considered `Index<usize>` before, but it would make this a lot easier.

cuviper:
One downside is that with my experiment of [custom index types](https://github.com/bluss/indexmap/pull/131#issuecomment-641638578), I would ideally replace this `Index<usize>` with `Index<Idx> for IndexMap<K, V, S, Idx>`, but I can't prove that `Index<Idx>` and `Index<&Q>` are distinct. We could still allow `Index<usize>` with any `Idx` though.

bluss:
I think we should just wait with this, in that case. If we have an indexing operator, IMO it must use the index type that we have parameterized the map with.

cuviper:
I brought this up in the forum: https://users.rust-lang.org/t/any-tricks-for-generic-overlap-with-references/45186

I now think the answer is that we shouldn't try to create generic `Index<Idx>` at all. Instead, we can just add concrete `Index<usize>` and `Index<u32>` for maps and sets of the corresponding index type. Then if a user has a newtype index, coherence will still let them write their own `Index<MyIndex>`, which just needs to call `get_index` and `unwrap`/`expect`.

bluss:
That makes sense but

> ith Index<Idx> for IndexMap<K, V, S, Idx>, but I can't prove that Index<Idx> and Index<&Q> are distinct

Won't we have a trait bound on `Idx`? If that's a local trait it should be enough, it will prove that no references implement it? (Or maybe that doesn't work because references are fundamental)

cuviper:
> (Or maybe that doesn't work because references are fundamental)

I think that's it. I shared a simplified playground, if you want to experiment with it:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=161bde1feeaa8c4aaaa864636f2f2ae1

```
error[E0119]: conflicting implementations of trait `std::ops::Index<&_>` for type `Container<_, &_>`:
[...]
   = note: downstream crates may implement trait `AsIndex` for type `&_`
```

cuviper:
@bluss any more thoughts on this? My plan for custom index types is to just implement each manually, rather than trying to have a blanket impl that *potentially* conflicts with `Index<&Q>`.

```rust
impl<K, V, S> Index<usize> for IndexMap<K, V, S, usize> {...}
impl<K, V, S> Index<u32> for IndexMap<K, V, S, u32> {...}
// perhaps all unsigned integers
// users can do their own `Index<NewTypeIndex>`
```

cuviper:
Let's go ahead with that plan then...

PR #133, opened 2020-06-23
Add FnvIndexMap, FnvIndexSet

konsumlamm:
Add type aliases for versions of `IndexMap` and `IndexSet` using a default FNV hasher, analog to those in the `fnv` crate itself. This adds an optional dependency to the `fnv` crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,10 +1,7 @@
 [package]
 name = "indexmap"
-version = "1.4.0"
-authors = [
-"bluss",
-"Josh Stone <cuviper@gmail.com>"
-]
+version = "1.4.1"
+authors = ["bluss", "Josh Stone <cuviper@gmail.com>"]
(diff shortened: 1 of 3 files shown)


bluss:
FxHasher and others are more interesting than this hasher, I'd lean towards declining this PR because it is a mere convenience, and it might not be worth the cost of a public dependency, even if that one is stable.

konsumlamm:
I understand, thanks for your response.

cuviper:
Maybe we could just document examples of this for a few hashers? Then we'd only need them as dev-dependencies for doc tests, and in fact we already have `fnv` for no-std testing.

konsumlamm:
> Maybe we could just document examples of this for a few hashers?

Sounds like a good idea. I'm closing this for now.

cuviper:
I opened #134 with a simple doc example.

PR #136, opened 2020-07-08
Add more methods matching the standard HashMap

cuviper:
- `pub fn get_key_value<Q: ?Sized>(&self, key: &Q) -> Option<(&K, &V)>`
  - Stabilized in Rust 1.40.
  - This is a middle-ground between `get` and our own `get_full` that
    includes the index. It's easy enough to add for compatibility.

- `pub fn remove_entry<Q: ?Sized>(&mut self, key: &Q) -> Option<(K, V)>`
  - Stabilized in Rust 1.27.
  - For `IndexMap`, this also adds `swap_`/`shift_remove_entry` choices.
  - These are also a middle-ground between (`swap_`/`shift_`)`remove` and
    our own `swap`/`shift_remove_full` that include the index. We don't
    have a plain `remove_full` though.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -342,6 +342,22 @@ where
         self.get_full(key).map(third)
     }
 
+    /// Return references to the key-value pair stored for `key`,
+    /// if it is present, else `None`.
+    ///
+    /// Computes in **O(1)** time (average).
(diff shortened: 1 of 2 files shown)


cuviper:
I'm going ahead with the merge, because there's no real API design here. This is just matching `std` methods, plus swap/shift semantics, and the implementations are trivial.

PR #137, opened 2020-07-20
Add support for amortized resizes

jonhoo:
Hi there!

As you may have [seen](https://twitter.com/jonhoo/status/1283852507056427008), over the past few weeks, I've been digging deep down into some latency spikes I was observing when using `IndexMap`. Further investigation revealed that they were caused by resizes, which can get quite expensive as the map grows large. To mitigate the issue, I built an amortized version of `hashbrown` ([griddle](https://github.com/jonhoo/griddle/)), and and amortized version of `Vec` ([atone](https://github.com/jonhoo/atone)), and then plugged them into `IndexMap`. This worked [really well](https://twitter.com/jonhoo/status/1284114860452777984), and cut my tail latency by several orders of magnitude.

I'm opening this mostly to serve as a place to discuss a path forward. It might be that that path forward is to publish a separate crate as I was told to do for `hashbrown` (https://github.com/rust-lang/hashbrown/pull/166#issuecomment-651386557). I'd be curious to hear your thoughts on including this as a proper feature of `indexmap`.

The changes in this PR are actually fairly few, but there were many changes needed just to be able to abstract away the use of `Vec`/slices to something that can be either a slice or an `atone::Vc` (this is the `EntryVec` type alias). The primary functional changes are:

 - Calls to `make_contiguous` prior to sorting, which are nightly only ([tracking issue](https://github.com/rust-lang/rust/issues/70929))
 - `impl Debug for IntoIter` does not print the items with `feature = "amortize"`, since `vec_deque::IntoIter` does not provide by-ref access to the "remainder"
 - Changes to `capacity` are a little less predictable (see the tests), since the storage is backed by a `VecDeque`, not a `Vec`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,12 +34,19 @@ bench = false
 autocfg = "1"
 [dependencies]
 serde = { version = "1.0", optional = true, default-features = false }
-rayon = { version = "1.0", optional = true }
+rayon_ = { version = "1.0", optional = true, package = "rayon" }
+atone = { version = "0.3.1", optional = true }
 
 [dependencies.hashbrown]
 version = "0.8.1"
 default-features = false
 features = ["raw"]
+
+[dependencies.griddle]
+version = "0.3.1"
+default-features = false
+features = ["raw"]
+optional = true
 
 [dev-dependencies]
 itertools = "0.9"
@@ -52,6 +59,10 @@ fxhash = "0.2.1"
 [features]
 # Serialization with serde 1.0
 serde-1 = ["serde"]
+rayon = ["rayon_", "atone/rayon"]
+
+# Use griddle over hashbrown, and atone over Vec, for amortized resizes
+amortize = ["griddle", "atone"]
 
 # for testing only, of course
 test_low_transition_point = []
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -86,9 +86,11 @@ extern crate alloc;
 extern crate std;
 
 #[cfg(not(has_std))]
+#[cfg_attr(feature = "amortize", allow(unused_imports))]
 use alloc::vec::{self, Vec};
 
 #[cfg(has_std)]
+#[cfg_attr(feature = "amortize", allow(unused_imports))]
 use std::vec::{self, Vec};
 
 #[macro_use]
@@ -112,6 +114,11 @@ pub use crate::map::IndexMap;
 pub use crate::set::IndexSet;
 
 // shared private items
+
+#[cfg(feature = "amortize")]
+type EntryVec<T> = atone::Vc<T>;
+#[cfg(not(feature = "amortize"))]
+type EntryVec<T> = Vec<T>;
 
 /// Hash value newtype. Not larger than usize, since anything larger
 /// isn't used for selecting position anyway.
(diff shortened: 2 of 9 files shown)


jonhoo:
@bluss and @cuviper, I'd be curious to hear your take on whether this is something that _might_ make it into `indexmap` (even if not in its current form), or whether you think I should just maintain a near-fork with this patch?

cuviper:
At a high level, I'm a little concerned about the global effect by doing this with a feature. There are trade-offs in this, and that would be okay if a root application crate made this choice (akin to a global allocator), but any crate in the dependency graph could activate this feature affecting `IndexMap` everywhere. From that angle, a forked variant seems better.

Cargo.toml
```diff
@@ -50,7 +57,11 @@ fxhash = "0.2.1"
 
 [features]
 # Serialization with serde 1.0
-serde-1 = ["serde"]
+serde-1 = ["serde", "atone/serde", "griddle/serde"]
+rayon = ["rayon_", "atone/rayon", "griddle/rayon"]
```
cuviper, line 61: I believe this means enabling either `serde-1` or `rayon` will always force `atone` and `griddle` to be built, regardless of `amortize`.

I don't think we really need all those transitive features anyway, just `atone/rayon`. Our `serde` support only needs plain iteration, and for `rayon` we only touch the vector.

cuviper: I think this is why CI failed on 1.36 thumb for `test-nostd`, because that enables `indexmap/serde-1`.

jonhoo:
> I'm a little concerned about the global effect by doing this with a feature

Yeah, that's a strong point. I guess the flip side is that keeping a separate crate would require ongoing maintenance to keep the projects in sync. I wonder if there's a way for `indexmap` to provide a `raw` module the same way `hashbrown` does, so that the fork doesn't have to also copy over all the internals... That would certainly make keeping a fork much less work! Though also potentially more work on your side.

Cargo.toml
jonhoo: Ah, yes, looks like you're right: https://github.com/rust-lang/cargo/issues/3494. And to quote from that:
> I don't believe it's possible to express what you're thinking here today with Cargo features.

But your point that we may not need those features is a good one — I'll play around with that. I think even for the vector, we _may_ be able to get away with not having the rayon feature on `atone` either.

cuviper:
Changes here are _usually_ pretty infrequent. The hashbrown change was a biggy, and I do have a few more things lined up, but after that I would expect it to slow down again. Plus, it's not like a fork needs to be aggressive about staying in sync, just good enough for your forked users' needs.

I'm not sure what a `raw` API would look like here...

jonhoo:
That's true. I guess I'll pursue a fork instead then. Can you give me an idea of the changes you have in mind?

As for a `raw` API, yeah, it'd be tricky. I can imagine what it'd look like (basically a "smart" `Bucket` wrapper whose methods maintain the `indexmap` invariants), basically like your current `map/core`, but I don't know if it'd even be sufficient for what I need. Specifically because the change from `[]` to `atone` is so invasive.

cuviper:
> Can you give me an idea of the changes you have in mind?

Direct indexing (#132), and also parameterizing the index type in the raw table (e.g. storing `u32` or a newtype index).

jonhoo:
Published as https://github.com/jonhoo/indexmap-amortized

cuviper:
Congrats! Not as clever as your previous names though... :wink:

jonhoo:
@cuviper hehe, I figured for this specifically it'd be good to keep the `indexmap` association.

One observation already: the amortize feature vs new crate decision keeps on occurring downstream. For example, I need amortization for [`evmap`](https://github.com/jonhoo/rust-evmap) and [`hashbag`](https://github.com/jonhoo/hashbag) too, and I _really_ don't want to maintain forks for _all_ of those. Even though it's awkward, I think I'm just going to go with a feature for this, and then hope that something like https://github.com/rust-lang/cargo/issues/2980 eventually provides me with a way to say that "this feature causes a divergence of behavior, and should produce a separate compilation target".

PR #142, opened 2020-07-24
Expand drain to accept any RangeBounds<usize>

cuviper:
Rather than just `RangeFull`, this lets us accept any `Range*` type or
even arbitrary `Bound` pairs, matching the signature of `Vec::drain`.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,4 +1,5 @@
 use core::iter::Enumerate;
+use core::ops::{Bound, Range, RangeBounds};
 
 pub(crate) fn third<A, B, C>(t: (A, B, C)) -> C {
     t.2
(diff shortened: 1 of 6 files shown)


src/util.rs
```diff
@@ -10,3 +11,29 @@ where
 {
     iterable.into_iter().enumerate()
 }
+
+pub(crate) fn simplify_range<R>(range: R, len: usize) -> Range<usize>
```
cuviper, line 15: I think this will also be useful for #104.

cuviper:
I'm going to go ahead and merge this -- the signature should be uncontroversial to match `Vec`, and it's not a breaking change since we still accept `RangeFull` as a possibility. We can always iterate on the implementation more later, if needed.

Issue #144, opened 2020-07-30
1.5.0 breaks h2 build


crackcomm:
After re-installing my environment I'm getting an error on same code base:

```sh
error[E0107]: wrong number of type arguments: expected 3, found 2
  --> /home/pah/.cargo/registry/src/github.com-1ecc6299db9ec823/h2-0.2.6/src/proto/streams/store.rs:15:10
   |
15 |     ids: IndexMap<StreamId, SlabIndex>,
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 3 type arguments

error: aborting due to previous error

For more information about this error, try `rustc --explain E0107`.
error: could not compile `h2`.
```

Is there a way to somehow set `has_std` for indexmap crate so deep in dependency tree?

Edit: It works on windows.

cuviper:
I suppose your lock file had an older version before, but do you have any idea what that was? You could try `cargo update -p indexmap --precise 1.4.0` or older, if the rest of your dependencies will allow it.

You say Windows works, so what is the target which fails? If that's using cross or a custom spec, it could be related to cuviper/autocfg#15.

Right now `has_std` is detected with an `autocfg` test, but we could add a manual feature to force that on.

crackcomm:
@cuviper that's exactly what happened, Cargo.lock was old.

Windows works and `stable` on Linux works as well.

Now a strange thing happened. I tried to build the repo with `stable` which failed because of unstable features in my code but `Checking h2 v0.2.6` passed. Once I switched back to `nightly` everything compiled just fine.

crackcomm:
I'd close this issue because I think it should be in autocfg repo. Thanks for help to you @cuviper and few people from the discord server.

cuviper:
Ultimately, whether the issue is here or in autocfg, I'm going to need a way to reproduce the failure. If you're doing something with with rustc flags that's not picked up by autocfg, that's the known issue, otherwise I don't know what's wrong.

crackcomm:
Fortunately, it was fresh Ubuntu on WSL initialized with a single script so you can reproduce this.

```sh
#!/bin/sh

sudo add-apt-repository -y ppa:avsm/ppa
sudo apt-get update &&
  sudo apt-get install -qy \
    m4 \
    apt-transport-https \
    ca-certificates \
    curl \
    gnupg-agent \
    software-properties-common \
    build-essential \
    git \
    wget patch diffutils zlib1g-dev libtinfo-dev \
[...]

PR #145, opened 2020-08-06
Add a feature to explicitly enable "std"

cuviper:
If you know you need "std", e.g. relying on default `S = RandomState`,
then you can enable this feature to force it on, bypassing target
detection. This may help scenarios where `autocfg` fails detection, like
issue #144, but at the very least it will speed up the build script by
avoiding extra `rustc` invocations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ matrix:
     - rust: stable
       env:
        - FEATURES='rayon'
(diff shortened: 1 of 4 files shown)


bluss (approved):

cuviper:
OK, I'm going to release this as 1.5.2.

PR #149, opened 2020-08-09
ci: make sure test commands are `&&`-chained

cuviper:
The new test command added in e0c627b1bea0f directly followed the prior
command, without `&&`-chaining the result. So if the tests before it
_failed_, we wouldn't notice...


cuviper:
Thankfully nothing has been failing!

PR #152, opened 2020-09-03
Update to hashbrown 0.9

mbrubeck:
Note: This increases the MSRV to Rust 1.36.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,7 @@ language: rust
 sudo: false
 matrix:
   include:
-    - rust: 1.32.0
-    - rust: 1.34.2
+    - rust: 1.36.0
     - rust: stable
       env:
        - FEATURES='serde-1'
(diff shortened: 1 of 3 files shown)


mbrubeck:
Note: If keeping the MSRV at 1.32 is a priority, hashbrown 0.9 could be changed to support that version, except when building with the optional `rayon` feature and the latest version of `crossbeam`.

cuviper:
I'm OK with raising MSRV with a minor version bump. My support threshold for a crate like this is about a year, and 1.36 was released 2019-07-04.

I was looking for interesting new features, and `const fn new` would be nice, but we'll need `Vec::new` too (Rust 1.39).

If you're itching for a release, go ahead and bump to 1.6.0 and I'll publish.

mbrubeck:
Bumped the minor version.

cuviper:
Thanks!

Issue #135, opened 2020-07-02
Possible changes for 2.0


cuviper:
Let's use this issue to discuss changes for a hypothetical 2.0.

### Proposed

- [x] Increase the MSRV from 1.18.
  - This is not strictly breaking, as we did state the intent to allow increases in minor releases, but in 2.0 we can do this without any fear of harming our dependents.
    - In particular, `serde_json` supports Rust 1.31.
  - #104 wants 1.28 for `RangeBounds`.
  - #131 wants 1.32 for `hashbrown`.
  - 1.34 adds `TryFrom`/`TryInto` which may be useful for custom indexing.
  - With 1.36, we could assume `alloc` support.
  - **DONE**: We went ahead with MSRV 1.32 in `indexmap` 1.5, and MSRV 1.36 in 1.6
- [ ] Change the default hasher to something faster.
  - e.g. `hashbrown` defaults to `ahash`.
  - Especially with MSRV 1.36, a `no_std` default would be nice.
  - We can wrap it to avoid committing to a specific public dependency.
- [ ] Parameterize the index type.
  - While we could try to default `Idx = usize`, this doesn't always work well with type inference.
- [ ] Maybe change the default `remove()`?
  - [(bluss)](https://github.com/bluss/indexmap/pull/131#issuecomment-651963896) I still think the current decision is fine, no change from my end for 2.0. (Good way to document this would be to say that "remove is the fastest way to remove an element, and if order should be preserved, ...")
- [x] `std` crate feature
- [ ] Consider _ordered_ `PartialEq + Eq`, more like `Vec` than `HashMap` (#153, #154).
  - This would also allow `PartialOrd`, `Ord`, and `Hash`.
- [x] `get_index_mut` should return `&K` instead of `&mut K` (https://github.com/bluss/indexmap/pull/160#issuecomment-736910427)

bluss:
Added item for a proper `std` crate feature. Makes it easier for deps

cuviper:
If we adopt MSRV 1.36 and a new default hasher, I think there will be no more need for `std` at all. But it's not certain that we should do that.

bluss:
I just have an attitude problem with ahash, would be more charmed by a pure-rust hash function.

But for an actual argument, for a public dependency we need them to have an 1.x version or equivalent.

ammkrn:
[...]

PR #154, opened 2020-09-08
map: Make Eq comparisons ordering-aware.

emilio:
This is a breaking change. Not sure if you really want to take this but
I think it's the right thing to do.

### DIFF
--- a/src/set.rs
+++ b/src/set.rs
@@ -820,7 +820,7 @@ where
     S2: BuildHasher,
 {
     fn eq(&self, other: &IndexSet<T, S2>) -> bool {
-        self.len() == other.len() && self.is_subset(other)
+        self.len() == other.len() && self.iter().zip(other.iter()).all(|(v1, v2)| v1 == v2)
     }
 }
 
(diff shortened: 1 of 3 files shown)


cuviper:
> This is a breaking change.

I've added it to #135 to consider for 2.0.

In the meantime, we could add a method for `ordered_eq`, but it's just:

```rust
map1.len() == map2.len() && map1.iter().eq(&map2)
```

emilio:
Yup, agree, `ordered_eq` is probably not quite worth it.

That seems reasonable, thank you!

emilio:
Do you want to leave this PR open? Or should I close it?

cuviper:
Let's close -- it will be here when we're ready for 2.0, but that's not really happening yet.

src/set.rs
```diff
@@ -1580,9 +1588,9 @@ mod tests {
         assert_eq!(&set_a & &set_b, empty);
         assert_eq!(&set_b & &set_a, empty);
         assert_eq!(&set_a | &set_b, set_c);
-        assert_eq!(&set_b | &set_a, set_c);
+        assert_ne!(&set_b | &set_a, set_c);
         assert_eq!(&set_a ^ &set_b, set_c);
-        assert_eq!(&set_b ^ &set_a, set_c);
+        assert_ne!(&set_b ^ &set_a, set_c);
```
bluss, line 1593: when this is revived later, this change does not preserve the intent of the test. It can of course be fleshed out when/if relevant later.

bluss:
Thanks for filing. Not a bug since it's the intentional design, but worth discussing for sure. I think a method for ordered eq would be worth it.

PR #157, opened 2020-09-16
Let feature "serde" work as well as "serde-1"

### LINKED ISSUE #84: Implement `serde` traits under `#[cfg(features = "serde")]`
Instead of `#[cfg(features = "serde-1")]` which is bit counter-intuitive and not in consistency with rust ecosystem.

cuviper:
The optional dependency on `serde` creates an implicit feature, but
we've only been enabling that implementation with `"serde-1"`. Now you
can directly enable `"serde"` to get full use, and `"serde-1"` is just
kept for compatibility.

If we ever need a `"serde-2"`, that can depend on package renaming.

Closes #84.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,7 +95,7 @@ use std::vec::{self, Vec};
 mod macros;
 mod equivalent;
 mod mutable_keys;
-#[cfg(feature = "serde-1")]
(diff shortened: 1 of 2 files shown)


bluss:
Makes sense

PR #158, opened 2020-09-18
Add serde_seq to serialize maps as a sequence 

### LINKED ISSUE #156: Order of IndexMap is not restored with some serde implementations
When using [`flexbuffers`](https://crates.io/crates/flexbuffers) the order of `IndexMap` is not restored on deserialize.

I am not sure if this should be considered a bug in `flexbuffers`, a bug in `indexmap` or a shortcoming of `serde`. [The implementation of `serse::Serialize for IndexMap`](https://github.com/bluss/indexmap/blob/master/src/serde.rs#L14) uses [`serialize_map`](https://docs.serde.rs/serde/trait.Serializer.html#tymethod.serialize_map) which does not specify if the keys should keep the order.

Maybe `IndexMap` should use `serialize_seq` instead, i.e. encode a map as an array of key-value pairs?

## Related issues
https://github.com/google/flatbuffers/issues/6120

## Reproduce
with `indexmap = { version = "=1.6.0", features = ["serde-1"] }`:

``` rust
use indexmap::IndexMap;
fn main() {
    let mut map = IndexMap::new();
    map.insert("b".to_owned(), 1_i32);
    map.insert("a".to_owned(), 0_i32);
    let flex = flexbuffers::to_vec(&map).unwrap();
    let back: IndexMap<String, i32> = flexbuffers::from_slice(&flex).unwrap();
    assert_eq!(map.keys().collect::<Vec<_>>(), back.keys().collect::<Vec<_>>());
}
```

This fails with

```
  left: `["b", "a"]`,
 right: `["a", "b"]`', src/main.rs:8:5
```

cuviper:
Our default serde implementation treats `IndexMap` as a normal map,
which is often a nice representation in serialization formats, but they
might not preserve the order.

This commit adds a `serde_seq` module with `serialize` and `deserialize`
functions, which makes it suitable for serde's field attributes, like
`#[serde(with = "indexmap::serde_seq")]`. This mode treats `IndexMap` as
a sequence of `(key, value)` pairs, which should always preserve order.

Closes #156.

### DIFF
--- a/src/serde_seq.rs
+++ b/src/serde_seq.rs
@@ -1,0 +1,112 @@
+//! Functions to serialize and deserialize an `IndexMap` as an ordered sequence.
+//!
+//! The default `serde` implementation serializes `IndexMap` as a normal map,
+//! but there is no guarantee that serialization formats will preserve the order
+//! of the key-value pairs. This module serializes `IndexMap` as a sequence of
+//! `(key, value)` elements instead, in order.
+//!
+//! This module may be used in a field attribute for derived implementations:
+//!
+//! ```
+//! # use indexmap::IndexMap;
+//! # use serde_derive::{Deserialize, Serialize};
+//! #[derive(Deserialize, Serialize)]
+//! struct Data {
+//!     #[serde(with = "indexmap::serde_seq")]
+//!     map: IndexMap<i32, u64>,
+//!     // ...
+//! }
+//! ```
+//!
+//! Requires crate feature `"serde"` or `"serde-1"`
+
+use serde::de::{Deserialize, Deserializer, SeqAccess, Visitor};
+use serde::ser::{Serialize, Serializer};
+
+use core::fmt::{self, Formatter};
+use core::hash::{BuildHasher, Hash};
+use core::marker::PhantomData;
+
+use crate::IndexMap;
+
+/// Serializes an `IndexMap` as an ordered sequence.
+///
+/// This function may be used in a field attribute for deriving `Serialize`:
+///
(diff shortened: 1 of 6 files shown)


src/serde_seq.rs
```diff
@@ -0,0 +1,112 @@
+//! Functions to serialize and deserialize an `IndexMap` as an ordered sequence.
+//!
+//! The default `serde` implementation serializes `IndexMap` as a normal map,
+//! but there is no guarantee that serialization formats will preserve the order
+//! of the key-value pairs. This module serializes `IndexMap` as a sequence of
+//! `(key, value)` elements instead, in order.
```
cuviper, line 6: I'm not sure whether serializing tuples is a good choice, although it's the easiest to implement. Alternatively, serde sequences may be heterogeneous, so we could use a flat sequence of `key0, value0, key1, value1, ...` instead. I don't know whether all serialization formats support that though.

cuviper: This is also just a helper, so we don't have to be perfect here. A third party can write a similar custom module with different choices, as there aren't issues like trait coherence for this.

cuviper: Yet another possibility is a pair of sequences, `([key, ...], [value, ...])`, but when deserializing we would have to read keys into temporary storage before we start reading values and insert into the map.

bluss (approved):

bluss:
This seems good to support, I fear most users won't find this feature until after the first bug happens, but at least there's a solution then.

PR #159, opened 2020-09-28
Reduce unsafe code with hashbrown 0.9.1

cuviper:


cuviper:
There are no public API changes to discuss here, and CI is green, so I'll go ahead and merge.

PR #160, opened 2020-09-30
Add more Vec/slice-like methods to maps and sets

cuviper:
```rust
impl<K, V, S> IndexMap<K, V, S> {
    pub fn truncate(&mut self, len: usize);
    pub fn split_off(&mut self, at: usize) -> Self where S: Clone;
    pub fn first(&self) -> Option<(&K, &V)>;
    pub fn first_mut(&mut self) -> Option<(&mut K, &mut V)>;
    pub fn last(&self) -> Option<(&K, &V)>;
    pub fn last_mut(&mut self) -> Option<(&mut K, &mut V)>;
    pub fn swap_indices(&mut self, a: usize, b: usize);
}

impl<T, S> IndexSet<T, S> {
    pub fn truncate(&mut self, len: usize);
    pub fn split_off(&mut self, at: usize) -> Self where S: Clone;
    pub fn first(&self) -> Option<&T>;
    pub fn last(&self) -> Option<&T>;
    pub fn swap_indices(&mut self, a: usize, b: usize);
}
```

I would prefer `&K` instead of `&mut K` in `IndexMap::first_mut` and
`last_mut`, but this is consistent with the existing `get_index_mut`.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -252,6 +252,13 @@ impl<K, V, S> IndexMap<K, V, S> {
         self.core.clear();
     }
 
+    /// Shortens the map, keeping the first `len` elements and dropping the rest.
+    ///
+    /// If `len` is greater than the map's current length, this has no effect.
+    pub fn truncate(&mut self, len: usize) {
+        self.core.truncate(len);
+    }
+
     /// Clears the `IndexMap` in the given index range, returning those
     /// key-value pairs as a drain iterator.
     ///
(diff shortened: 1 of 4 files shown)


bluss:
`get_index_mut` seems to be an oversight. Mutable key access is supposedly hidden away in the trait `MutableKeys` and I think it seems ok to continue that.

src/map.rs
```diff
@@ -693,6 +717,22 @@ impl<K, V, S> IndexMap<K, V, S> {
         self.as_entries_mut().get_mut(index).map(Bucket::muts)
     }
 
+    pub fn first(&self) -> Option<(&K, &V)> {
+        self.as_entries().first().map(Bucket::refs)
+    }
+
+    pub fn first_mut(&mut self) -> Option<(&mut K, &mut V)> {
```
bluss, line 724: If you agree I'd do &K here, that's the initial interface we were going for in this crate (while still allowing opt-in access to mutable keys with the trait).

bluss (approved):
Comment about &K applies

bluss (changes requested):
I guess I technically can't approve while there are still methods without doc comment, but I like the idea.

cuviper:
OK, I added more docs for the public methods.

At some point we should probably go for `#![deny(missing_docs)]`, because there are a lot more missing, but I don't want to force that into this PR.

cuviper:
Oh, I switched to `&K` too -- force-pushed to split up that commit.

bluss (approved):

PR #161, opened 2020-11-25
Implement functionality to shift_remove in bulk

philippeitis:
Currently, there aren't any good options for removing multiple indices and maintaining ordering - to do so, you'd either do

```rust
    fn remove_items(&mut self, indices: Vec<usize>) {
        for index in indices {
            self.items.shift_remove_index(index);
        }
    } 
```

or

```rust
    fn remove_items(&mut self, indices: Vec<usize>) {
        for index in indices {
            self.items.remove_index(index);
        }
        self.items.sort();
    }
```

which would take O(nm) time and O(m+nlogn) time respectively, but this is pretty terrible when your IndexMap is large (eg. at 50k elements and 200 keys, you'd do almost 5 million individual shifts on average).

This implements a version which operates on a mutable array of indices (as the indices need to be sorted to go from smallest to largest). This should reduce the runtime to O(n +mlogm) for any amount of insertions, which is a massive improvement. However, it might be a bit slower than shift_remove_finish for small removals - doing a full sweep isn't possible, since identifying the correct amount to shift elements by requires a binary search on each index.

I haven't implemented it for `shift_remove_full`, since this would likely require allocating space for the indices so that they can be sorted.

This PR would allow 

```rust
    fn remove_items(&mut self, indices: &mut [usize]) {
        self.items.remove_indices(&mut indices);
    }
```

It might also be useful to allow removing ranges - but it seems that similar methods (eg. getting a slice or mutable slice of elements) are blocked from being implemented at the moment.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -718,6 +718,20 @@ impl<K, V, S> IndexMap<K, V, S> {
     pub fn shift_remove_index(&mut self, index: usize) -> Option<(K, V)> {
         self.core.shift_remove_index(index)
     }
+
+    /// Remove the key-value pairs by the provided indices.
+    ///
+    /// Valid indices are *0 <= index < self.len()*
+    ///
+    /// Like `Vec::remove`, the pairs are removed by shifting all of the
+    /// elements that follow, preserving their relative order.
+    /// **This perturbs the index of all of those elements!**
+    ///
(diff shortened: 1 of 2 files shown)


cuviper:
Yes, I think `retain` is the right call here.

PR #162, opened 2020-12-01
Fix outdated docs

taiki-e:
[Seems the current MSRV is 1.36](https://github.com/bluss/indexmap/pull/152), but some docs still say it's 1.32.


cuviper:
Thanks!

PR #164, opened 2020-12-01
Migrate CI from travis to github actions

### LINKED ISSUE #163: Migrate travis-ci to github actions
I could do this (I've started some of my other repos). What do you think? It seems to be the way to go (travis-ci.com being uncertain).

bluss:
Drop twox-hash from test-nostd to fix the builder for Rust 1.36, just use a dummy hasher.

CI should now be really speedy. Added clippy run (it can be disabled/ignored if it's annoying).

Old travis config was removed - travis-ci.org is EOL at the end of this year (there would be a
possibility of migrating to travis-ci.com but conditions for free CI are uncertain).
Thanks to travis-ci.org for their service until now.

Fixes #163


cuviper:
LGTM, but you should change/remove the build status badge in the README too.

PR #170, opened 2021-01-20
Change OccupiedEntry::key() to return the existing key in the map

### LINKED ISSUE #169: OccupiedEntry::key() gives the new key instead of the old one
```rust
struct Item { value: i32, id: &'static str }

impl PartialEq for Item {
    fn eq(&self, other: &Item) -> bool { self.value == other.value }
}
impl Eq for Item {}
impl std::hash::Hash for Item {
    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
        self.value.hash(state);
    }
}


fn main() {
    use indexmap::map::{IndexMap as Map, Entry};

    let mut map = Map::new();
    map.insert(Item { value: 3, id: "old" }, 200);
    
    match map.entry(Item { value: 3, id: "new" }) {
        Entry::Vacant(e) => panic!(),
        Entry::Occupied(e) => println!("{}", e.key().id),
    }
}
```
**output:**
```
new
```

Whereas if you instead `use std::collections::hash_map::{HashMap as Map, Entry};` then it prints `old`.  This difference in behavior seems surprising.

cuviper:
The new behavior is consistent with `HashMap` entries in the standard
library, and it has been our intent to match that API when possible.

Fixes #169.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1649,6 +1649,28 @@ mod tests {
     }
 
     #[test]
+    fn occupied_entry_key() {
+        // These keys match hash and equality, but their addresses are distinct.
+        let (k1, k2) = (&mut 1, &mut 1);
+        let k1_ptr = k1 as *const i32;
+        let k2_ptr = k2 as *const i32;
+        assert_ne!(k1_ptr, k2_ptr);
+
+        let mut map = IndexMap::new();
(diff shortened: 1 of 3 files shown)


bluss (approved):
Documentation 💯

Issue #173, opened 2021-02-21
Feature request: shift_insert


TethysSvensson:
I would like to be able to insert an element at a particular index in the indexmap (knowning full well, that this will cause `O(n)` move operations).

cuviper:
I suppose if the item already exists at any index, you want it moved to the new location?

Rough sketch:

```rust
    pub fn insert_at(&mut self, index: usize, key: K, value: V) -> Option<V> {
        let (i, old) = self.insert_full(key, value);
        if i < index {
            // TODO: adjust `indices` too
            self.entries[i..=index].rotate_left(1);
        } else if i > index {
            // TODO: adjust `indices` too
            self.entries[index..=i].rotate_right(1);
        }
        old
    }
```

The crude way to update `indices` is `IndexMapCore::rebuild_hash_table` -- reinserting everything, but at least we have saved hashes. We can probably do better though, more like the heuristics in `shift_remove`.

TethysSvensson:
> I suppose if the item already exists at any index, you want it moved to the new location?

For my particular use-case I wish to keep the item at the current index, but I need to modify it as well. However, performance is not very critical at all, so I am completely okay with handling this by doing a second lookup.

I think you are right, that the most reasonable behavior would be to move the item it.

I guess the way to completely support my use-case would be to add `shift_insert` methods to both `VacantEntry` and `OccupiedEntry`, but I am not sure if this is worth the added complexity.

cuviper:
Please see #176 for a `move_index` method that I think would be useful to you. That does the hard part of updating indices, and then we can possibly build new insertion or entry methods on top of that.

PR #83, opened 2018-11-19
Additional operations for manipulating iteration order

allen-marshall:
This PR adds operations to `IndexMap` (and `IndexSet`) that allow an entry to be moved to a specific position in the iteration order, while preserving the order of all other entries. The operation's average computation time is proportional to the distance between the entry's current position and the new position, assuming that probing the hash table takes constant time on average.

The PR requires an upgrade to Rust version 1.26, because it relies on the `rotate_left` and `rotate_right` methods of the slice type. If upgrading to 1.26 is deemed inappropriate at this time, it may be necessary to delay merging this PR.

The new operations also may allow for an alternative implementation of `ordered_remove` (from PR #15), in which the element being removed is first moved to the end of the ordering, then `IndexMap::pop()` is called. This might be more efficient if the element to be removed is already close to the end. I haven't included that idea in the PR, but it could be considered if the PR is accepted.

Please let me know if you find any problems in the PR that I need to fix. One thing I am unsure about is whether returning `Option<()>` is or is not an appropriate/idiomatic way of supporting error detection in a method that wouldn't normally return anything.

### Use Case
My use case for this PR is that I am planning to create a GUI application where the user can manage a list of items, including reordering the items by dragging and dropping an item in the list. Internally, the items may reference each other by a unique key. Ideally, I want to be able to look up an item efficiently by both its numerical index and its unique key. IndexMap seems like a good fit for this use case, and this PR would make it easier for me to implement the reordering through drag and drop.

PR #176, opened 2021-02-27
Add move_index to change the position of an entry

cuviper:
This moves the position of a key-value pair from one index to another by
shifting all other pairs in-between, making this an O(n) operation.

This could be used as a building-block for other operations, like #173
which wants to insert at a particular index. You can `insert_full` to
insert it _somewhere_, then choose whether to `move_index` depending on
whether you want to also change pre-existing entries.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -19,6 +19,10 @@
   - The new `IndexMap::shrink_to` and `IndexSet::shrink_to` methods shrink
     the capacity with a lower bound.
 
+  - The new `IndexMap::move_index` and `IndexSet::move_index` methods change
+    the position of an item from one index to another, shifting the items
+    between to accommodate the move.
+
   - The new `map::Slice<K, V>` and `set::Slice<T>` offer a linear view of maps
     and sets, behaving a lot like normal `[(K, V)]` and `[T]` slices. Notably,
     comparison traits like `Eq` only consider items in order, rather than hash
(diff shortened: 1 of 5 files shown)


cuviper:
I'm not entirely committed to the name `move_index`. It's somewhat similar to `swap_indices`, but I chose singular `index` because the movement of the one `from` index is the main point, while all the other shifted indices are a side-effect. Semantically, this is equivalent to `let value = vec.remove(from); vec.insert(to);`, or a slice rotation as this implementation actually does in its `entries`. Maybe `shift` or `rotate` would be better in the name.

cuviper:
I just noticed that #83 had a similar method called `reorder_entry_index`.

praveenperera:
I would love to get access to this function. Still thinking about merging it in?

cuviper:
Sure -- do you have any thoughts about the name?

praveenperera:
I think move_index works best. 

Reads really nicely as move index from x to y

praveenperera:
I tried the function out on my [fork](https://github.com/praveenperera/indexmap). 

Visually it sometimes does not work in an intuitive way. For example if you move the the item to the second item in the set, it actually moves it to the end. Also if you're moving it an index over, it moves two spaces down. The function is still useful though. 

I just ended up doing: 

1. `shift_remove`
2. `insert`
3. `move` index

It's possible I was using it wrong, it's been a long day. I will confirm tomorrow with a test. But thought I'd mention it.

praveenperera:
Ignore my last comment if you saw it (now deleted). This works great, I tired it out on my fork. This is exactly what I needed to make IndexSet and IndexMap work with my data structure.

PR #179, opened 2021-03-15
Prefer copied() iterators instead of cloned()

cuviper:
`Copied` is available since Rust 1.36 and Rayon 1.2.


bluss (approved):
Nice. The only non-test impact is IndexSet::extend (just clerical note)

cuviper:
Yeah -- it *might* have some codegen benefit in `extend`, but I wouldn't expect much. We don't need to release-note this IMO.

PR #185, opened 2021-04-24
Fix typo in map.rs

taiki-e:


cuviper:
It's amazing how well "the the" can hide in plain sight. Thanks!

PR #187, opened 2021-05-06
Document average time complexity of ```get_index_of()``` method

Kaduo:


bluss:
Thanks!

your-diary:
I think it would be great if `IndexSet::get_index_of()` is also documented.
(This PR only documents `IndexMap::get_index_of()`.)

PR #188, opened 2021-05-11
Simplify logic around `#[cfg(has_std)]`

bjorn3:
Unconditionally using `extern crate alloc;` allows removing several `#[cfg(has_std)]`/`#[cfg(not(has_std))]`.


cuviper:
Thanks! It looks like all remaining conditionals depend on `RandomState`, so that's as far as we can go.

PR #190, opened 2021-06-07
update hashbrown to 0.11 and rayon to 1.5

xMAC94x:


bluss:
See the other open PR

PR #195, opened 2021-07-30
Add IndexMap::into_{keys,values}

taiki-e:
This adds equivalents of the same methods of HashMap and BTreeMap stabilized in Rust 1.54.

- https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.into_keys
- https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.into_values
- https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.into_keys
- https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.into_values

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -162,6 +162,9 @@ impl<K, V> Bucket<K, V> {
     fn key(self) -> K {
         self.key
     }
+    fn value(self) -> V {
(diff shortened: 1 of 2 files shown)


cuviper:
Looks good, thanks!

PR #196, opened 2021-07-30
Iterator polish

cuviper:
This is a variety of small improvements for the iterators:

- Use `double_ended_iterator_methods!`
- Specialize `nth_back` just like `nth`
- Add `rfold` for iterators built on `Chain`
- `impl ExactSizeIterator for Drain`
- `impl FusedIterator` for all serial iterators
- `impl Debug for Drain`, `ParIterMut`, and `ParValuesMut`

PR #198, opened 2021-08-17
Fix missing `ExactSizeIterator` impl for `Drain`

Ten0:


cuviper:
Thank you, but I already have this as part of #196.

cuviper:
And now that's merged. Thanks for your attention anyway!

Ten0:
Perfect, thanks!

PR #203, opened 2021-09-16
Make with_hasher const

### LINKED ISSUE #202: Create indexmap/set in static contexts
Would it be possible to add a `new`-like function that is `const`? It would be useful to create static maps without the aid of something like `once_cell`.

From scanning the code, it seems like `with_hasher` is maybe the right function to make `const`, since you don't have to make a hasher (it's given) and "no allocation" is already in the spec for that function.

mwillsey:
Closes #202 

If you need `hash_builder` that can be made in the static context, consider [`ahash`](https://docs.rs/ahash/*/ahash/struct.RandomState.html#method.with_seeds).

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -166,10 +166,7 @@ impl<K, V, S> IndexMap<K, V, S> {
     #[inline]
     pub fn with_capacity_and_hasher(n: usize, hash_builder: S) -> Self {
         if n == 0 {
-            IndexMap {
-                core: IndexMapCore::new(),
-                hash_builder,
-            }
+            Self::with_hasher(hash_builder)
         } else {
             IndexMap {
                 core: IndexMapCore::with_capacity(n),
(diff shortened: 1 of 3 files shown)


src/map.rs
```diff
@@ -179,8 +176,11 @@ impl<K, V, S> IndexMap<K, V, S> {
     }
 
     /// Create a new map with `hash_builder`
```
cuviper, line 178: Can you add a note about the `const`ness, and that (like `new`) it does not allocate?
(ditto on the set)

mwillsey: Done. I also noted `ahash` in particular because not many hash builders are const-able.

src/map.rs
```diff
@@ -178,9 +175,17 @@ impl<K, V, S> IndexMap<K, V, S> {
...
 
-    /// Create a new map with `hash_builder`
-    pub fn with_hasher(hash_builder: S) -> Self {
-        Self::with_capacity_and_hasher(0, hash_builder)
+    /// Create a new map with `hash_builder`.
+    ///
+    /// This function is `const`, so it
+    /// can be called in `static` contexts.
+    /// If you need `hash_builder` that can be made in the static context,
+    /// consider [`ahash`](https://docs.rs/ahash/0.7.4/ahash/struct.RandomState.html#method.with_seeds).
```
bluss, line 183: it's nice to help the user but I'd like to avoid having such versioned urls - they are stale quickly - and fast-changing information - here. I would suggest we skip mentioning Ahash.

mwillsey: done!

bluss (approved):
Cool, looks good to me

cuviper (approved):

PR #205, opened 2021-10-21
Add `IndexMap::from(array)` and `IndexSet::from(array)`

rouge8:
This patch adds `IndexMap::from(array)` and `IndexSet::from(array)` to match the API for `HashMap`, `HashSet`, etc. as of https://github.com/rust-lang/rust/pull/84111.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -4,5 +4,6 @@ fn main() {
         Some(_) => autocfg::emit("has_std"),
         None => autocfg::new().emit_sysroot_crate("std"),
     }
+    autocfg::new().emit_rustc_version(1, 51);
     autocfg::rerun_path("build.rs");
 }
--- a/src/map.rs
+++ b/src/map.rs
@@ -1345,6 +1345,25 @@ where
     }
 }
 
+#[cfg(has_std)]
+impl<K, V, const N: usize> From<[(K, V); N]> for IndexMap<K, V, RandomState>
+where
+    K: Hash + Eq,
+{
+    /// # Examples
+    ///
+    /// ```
+    /// use indexmap::IndexMap;
+    ///
+    /// let map1 = IndexMap::from([(1, 2), (3, 4)]);
+    /// let map2: IndexMap<_, _> = [(1, 2), (3, 4)].into();
+    /// assert_eq!(map1, map2);
+    /// ```
+    fn from(arr: [(K, V); N]) -> Self {
+        std::array::IntoIter::new(arr).collect()
+    }
+}
+
 impl<K, V, S> Extend<(K, V)> for IndexMap<K, V, S>
 where
     K: Hash + Eq,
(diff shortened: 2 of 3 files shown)


rouge8:
@bluss bumping this now that CI is passing

cuviper (changes requested):

build.rs
```diff
@@ -4,5 +4,6 @@ fn main() {
         Some(_) => autocfg::emit("has_std"),
         None => autocfg::new().emit_sysroot_crate("std"),
     }
+    autocfg::new().emit_rustc_version(1, 51);
```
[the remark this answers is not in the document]
rouge8: Should I bump it in this PR?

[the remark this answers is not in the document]
cuviper: Well, I just looked at hashbrown and they may be bumping in the near future to MSRV 1.56, which we should probably follow when we do our bump. So... let's delay that decision for now.

src/map.rs
```diff
@@ -1340,6 +1340,25 @@ where
     }
 }
 
+#[cfg(all(has_std, rustc_1_51))]
```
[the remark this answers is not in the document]
rouge8: done!

src/map.rs
```diff
@@ -1342,6 +1342,25 @@ where
     }
 }
 
+#[cfg(all(rustc_1_51))]
```
cuviper, line 1345:
```suggestion
#[cfg(rustc_1_51)]
```

src/map.rs
```diff
@@ -1836,4 +1855,15 @@ mod tests {
         assert!(values.contains(&'b'));
         assert!(values.contains(&'c'));
     }
+
+    #[test]
+    #[cfg(all(rustc_1_51))]
```
cuviper, line 1860:
```suggestion
    #[cfg(rustc_1_51)]
```

src/set.rs
```diff
@@ -840,6 +840,25 @@ where
     }
 }
 
+#[cfg(all(rustc_1_51))]
```
cuviper, line 843:
```suggestion
#[cfg(rustc_1_51)]
```

src/set.rs
```diff
@@ -1707,4 +1726,13 @@ mod tests {
         assert_eq!(&set_c - &set_d, set_a);
         assert_eq!(&set_d - &set_c, &set_d - &set_b);
     }
+
+    #[test]
+    #[cfg(all(rustc_1_51))]
```
cuviper, line 1731:
```suggestion
    #[cfg(rustc_1_51)]
```

cuviper (changes requested):
Just some cleanup on the `cfg`s.

rouge8:
> Just some cleanup on the `cfg`s.

Oops, good catch. Fixed!

cuviper:
Oh, we do still need `std` because of `RandomState` -- that's too bad.

rouge8:
> Oh, we do still need `std` because of `RandomState` -- that's too bad.

Ah okay. Reverted removing `has_std`.

cuviper (approved):
Thanks!

PR #206, opened 2021-11-12
IndexMultimap

rinde:
This is a multimap implementation using `IndexMap` and `IndexSet`. Would this be a good fit in this repository?

If the answer..
- ..is yes, I will further improve `IndexMultimap` such that it's functionality/methods is on par with `IndexMap` and `IndexSet`. I'm also open to hearing any suggestions for improvements.
- ..is no, I will most likely put this in a separate crate.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -96,6 +96,7 @@ pub mod serde_seq;
 mod util;
 
 pub mod map;
+pub mod multimap;
 pub mod set;
 
 // Placed after `map` and `set` so new `rayon` methods on the types
--- a/src/multimap.rs
+++ b/src/multimap.rs
@@ -1,12 +1,13 @@
+use core::iter::FromIterator;
 use std::borrow::Borrow;
 use std::collections::hash_map::RandomState;
 use std::hash::BuildHasher;
 use std::hash::Hash;
 use std::iter::repeat;
 
-use indexmap::Equivalent;
-use indexmap::IndexMap;
-use indexmap::IndexSet;
+use crate::equivalent::Equivalent;
+use crate::IndexMap;
+use crate::IndexSet;
 
 /// Index map with multiple (unique) values per key.
 ///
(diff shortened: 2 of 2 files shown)


cuviper:
Is there anything in this that will require crate-local access to `IndexMap`? I don't see anything so far. So I think having it in a separate crate will be fine, and it will allow you to iterate faster.

The design choice of an inner `IndexSet` is also strange to me. Most multimaps I've seen place no constraints on the value type, and just collect them as an opaque bag (probably a vector). In cases where you really do want de-duplication and fast value lookups, a set makes sense -- but will you expose any "index" capabilities from that, or could it just be a regular `HashSet`?

I think any of those choices could be correct for different domains, but that's more reason not to "bless" one in this crate.

rinde:
Thanks for the quick reply.

> Is there anything in this that will require crate-local access to IndexMap?
No, it only builds directly on top of it. And I think it makes somewhat sense that `IndexMultimap` lives close to `IndexSet` & `IndexMap` to form a family of data structures that have deterministic iteration order.

The choice to use a set was born out of the requirements for what we needed in our application. The reason that it needed to be an `IndexSet` specifically was that our application also requires deterministic iteration order.

I see your point that this particular multimap implementation makes specific implementation choices that may make sense in certain domains but not in others. I will look into some options to make this implementation more generic/configurable such that it is more broadly applicable.

Thanks for the feedback!

rinde:
Short update on this topic, I continued working on this implementation in https://github.com/rinde/more_collections and now have the following implementations:
 - `HashSetMultimap` that works similar to `HashMap<K, HashSet<V>>`
 - `HashVecMultimap` that works similar to `HashMap<K, Vec<V>>`
 - `IndexSetMultimap` that works similar to `IndexMap<K, IndexSet<V>>`
 - `IndexVecMultimap` that works similar to `IndexMap<K, Vec<V>>`

cuviper:
Hi, thanks for the update! Given that you have been publishing that crate for a while as well, I think we can close it here.

PR #208, opened 2021-11-21
Clarify docs for `.pop()`

JakobDegen:
Most other methods specify how they affect insertion order; this PR adds those notes to `.pop()` as well.


bluss:
Thanks

PR #211, opened 2021-12-04
feat: add unstable sorting methods

### LINKED ISSUE #209: Add Unstable Sort Methods
It would be nice to have access to the unstable sorting algorithms:

- `IndexMap::sort_unstable_keys`
- `IndexMap::sort_unstable_by`
- `IndexMap::sorted_unstable_by`
- `IndexSet::sort_unstable`
- `IndexSet::sort_unstable_by`
- `IndexSet::sorted_unstable_by`

I can do a PR if there's interest.

bhgomes:
Closes #209.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -664,18 +664,18 @@ where
 
     /// Sort the map’s key-value pairs by the default ordering of the keys.
     ///
-    /// See `sort_by` for details.
+    /// See [`sort_by`](Self::sort_by) for details.
     pub fn sort_keys(&mut self)
     where
         K: Ord,
     {
-        self.with_entries(|entries| {
-            entries.sort_by(|a, b| Ord::cmp(&a.key, &b.key));
+        self.with_entries(move |entries| {
+            entries.sort_by(move |a, b| K::cmp(&a.key, &b.key));
         });
     }
 
     /// Sort the map’s key-value pairs in place using the comparison
-    /// function `compare`.
+    /// function `cmp`.
     ///
     /// The comparison function receives two key and value pairs to compare (you
     /// can sort by keys or values or their combination as needed).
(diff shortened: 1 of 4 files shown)


cuviper:
How about `rayon/set.rs` too?

src/map.rs
```diff
@@ -706,6 +706,52 @@ where
...
+        });
+    }
+
+    /// Sort the map's key-value pairs in place using the comparison function `cmp`, but
+    /// may not preserve the order of equal elements.
+    ///
+    /// The comparison function receives two key and value pairs to compare (you
+    /// can sort by keys or values or their combination as needed).
+    ///
+    /// Computes in **O(n log n + c)** time and **O(n)** space where *n* is
```
cuviper, line 728: I think we can drop the part about space -- unstable sort is in-place, with no additional memory allocation.

src/set.rs
```diff
@@ -580,7 +580,41 @@ where
...
+    pub fn sort_unstable(&mut self)
+    where
+        T: Ord,
+    {
+        self.map.sort_unstable_keys()
+    }
+
+    /// Sort the set's values in place using the comparison funtion `cmp`.
+    ///
+    /// Computes in **O(n log n)** time and **O(n)** space. The sort is unstable.
```
cuviper, line 599: Ditto, no extra space is used.

cuviper (changes requested):
Looks good, just a nit about documenting extra space or lack thereof.

src/map.rs
cuviper:
```suggestion
    /// Computes in **O(n log n + c)** time where *n* is
```

src/set.rs
cuviper:
```suggestion
    /// Computes in **O(n log n)** time. The sort is unstable.
```

PR #212, opened 2021-12-09
Add minimal support for rustc-rayon

cuviper:
This adds basic `IntoParallelIterator` support for the `rustc-rayon` fork of `rayon`, as needed by rust-lang/rust#90842 (cc @pierwill). The optional dependency is commented as "internal" for `rustc`'s use alone, and we should feel free to make breaking changes like semver-bumps in coordination with compiler updates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
           - rust: stable
             features: rayon
           - rust: stable
+            features: rustc-rayon
+          - rust: stable
             features: std
           - rust: beta
             features:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,10 @@ autocfg = "1"
 serde = { version = "1.0", optional = true, default-features = false }
 rayon = { version = "1.2", optional = true }
 
+# Internal feature, only used when building as part of rustc,
+# not part of the stable interface of this crate.
+rustc-rayon = { version = "0.3", optional = true }
+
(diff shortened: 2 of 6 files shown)


bluss (approved):
looks good to me, seems sensible.

bluss:
This makes me wonder about more hacky ideas, could rustc just patch in an alternative fork of rayon from the top of its cargo projects? I don't think the cargo patch feature can do exactly that, at least not without requiring version/crate name to stay identical.

cuviper:
Cargo patches affect the whole workspace, and I'm not sure we would *want* to force that fork everywhere since it lags behind. Really, we should do away with that fork, but that requires more work and nobody is really using the parallel rustc anyway...

bluss:
Yes I guess the fork would have to be a lot more careful about exactly what it's changing if it should work, it would need to be designed for being used that way.

pierwill:
This is amazing! Thank you @cuviper. :)

pierwill:
@cuviper: does this support `IntoParallelIterator` for `IndexSet` and `IndexMap` both?

cuviper:
Yes, this has `IntoParallelIterator` for `IndexMap`, `&IndexMap`, `&mut IndexMap`, `IndexSet`, and `&IndexSet`.

cuviper:
Confirmed on zulip that it is working with the "rustc-rayon" feature properly enabled. :)

I'm merging this, then I'll check other pending PRs and see about publishing a release.

PR #218, opened 2022-02-04
Add miri builder

erickt:
This adds a builder to run `cargo miri` in order to check for certain classes of undefined behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,3 +93,18 @@ jobs:
           components: clippy
       - run: cargo clippy
 
+  miri:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - nightly
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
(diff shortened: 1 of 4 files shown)


cuviper:
Any idea about that failure? I'm running locally and it's okay so far, but miri is also up to 10GB memory after 13 minutes. Maybe CI went OOM? Anyway, as long as there's no real issue there, we should probably `#[cfg(not(miri))]` some things so we don't inflate the CI time so much.

erickt:
I think we're tripping over https://github.com/rust-lang/miri/issues/1367, where the stacked borrow check is superlinear. We probably need to reduce the cycles for miri. I'll switch some of these expensive tests to that.

One interesting thing to note though that for [map::tests::insert_2](https://github.com/bluss/indexmap/blob/master/src/map.rs#L1565) uses about 2.2GB, but swapping in a HashMap, it only uses 338MB.

erickt:
I pushed up a new version that uses lower limits when run inside miri. Hopefully that'll use less memory and run faster.

cuviper:
Thanks! The new 4m miri CI time is totally reasonable.

PR #219, opened 2022-02-09
Prepare for indexmap 2.0

cuviper:
Of the list in #135, the ones addressed here are:

- Increase MSRV to 1.56
- Always use an explicit feature for `std`
- Make `get_index_mut` return `&K`

I also added a few new features:

- Add `MutableKeys::get_index_mut2`
- `impl Debug for IterMut` and `ValuesMut`
- Add `shrink_to`


cuviper:
Regarding other points from #135:

- Change the default hasher -- I think we should still consider this if we choose something `no_std` and hide it in a newtype. I don't think it's _necessary_ though, especially now that we have a cleaner `std` story.
- Parameterize the index type -- that would be #147, and I would like to see that now, or else abandon the idea. It can be rebased and merged after this PR though.
- Maybe change the default `remove()`? I think we're ok as-is.
- Consider _ordered_ `PartialEq + Eq` -- now is the time to talk about it, but I'm inclined to leave this alone. But we could merge #177 which has that behavior in slices.

bluss:
Awesome. Thanks a lot, and sorry for the long delay

cuviper:
Alright, I've rebased, and I push an `indexmap-1.x` branch for older stable work. Let's merge and see about those other features separately!

PR #222, opened 2022-03-27
Add a `replace_full` method on `IndexSet`

### LINKED ISSUE #221: Implement a `replace_full` method on `IndexSet`
Hello! I was wondering whether it would be possible to implement a `replace_full` method on `IndexSet`? Similarly to how there are the `insert` and `insert_full` methods, I think it might also be useful to also have a `replace_full` method.

```rust
pub fn replace_full(&mut self, value: T) -> (usize, Option<T>)
```

I believe the semantics would be:
- If the item already exists, replace it with `value` and return its index (which will remain unchanged) as well as its old value.
- Otherwise, insert the item into the set and return its index as well as `None`.

zakcutner:
* Add a new `replace_full` method, which behaves like `replace` but also
  returns the index of the item.

* Clarify that `replace` and `replace_all` do not modify the replaced
  item's insertion order in the documentation.

* Add test coverage for `replace` and `replace_full` by copying the
  existing tests for `insert` and `insert_full`.

Closes #221

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "indexmap"
 edition = "2018"
-version = "1.8.0"
+version = "1.8.1"
 authors = [
 "bluss",
 "Josh Stone <cuviper@gmail.com>"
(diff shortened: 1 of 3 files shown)


cuviper (approved):
Looks good, thanks!

I'll go ahead with a release bump in this PR too...

zakcutner:
Great, thanks so much!

PR #239, opened 2022-08-29
Add Natvis visualizations for core `indexmap` types

ridwanabdillahi:
This change adds Natvis visualizations for some `indexmap` types to help improve the debugging experience on Windows.

Natvis is a framework that can be used to specify how types should be viewed under a supported debugger, such as the Windows debugger (WinDbg) and the Visual Studio debugger.

When debugging types such as `IndexMap` and/or `IndexSet` under a Windows debugger, the entries for these data structures are not always clear upon first glance. The internal data structure is shown and it's not presented in a user friendly way. The Rust compiler does have Natvis support for some types, but this is limited to some of the core libraries and not supported for external crates.

https://github.com/rust-lang/rfcs/pull/3191 proposes adding support for embedding debugging visualizations such as Natvis in a Rust crate. This RFC has been approved, merged and implemented.

This PR adds:

Natvis visualizations for some of the core types such as `IndexMap` and `IndexSet`.
Tests for testing visualizers embedded in the `indexmap` crate.
Updates to the CI pipeline to ensure tests for visualizers are run so they do not break silently.
A new `debugger_visualizer` feature for the `indexmap` crate to enable the unstable `debugger_visualizer` Rust feature.

### DIFF
--- a/debug_metadata/indexmap.natvis
+++ b/debug_metadata/indexmap.natvis
@@ -1,0 +1,29 @@
+<AutoVisualizer xmlns="http://schemas.microsoft.com/vstudio/debugger/natvis/2010">
+  <Type Name="indexmap::set::IndexSet&lt;*,*&gt;">
+    <DisplayString>{{ len={map.core.entries.len} }}</DisplayString>
+    <Expand>
+      <ExpandedItem>map</ExpandedItem>
+    </Expand>
+  </Type>
+  <Type Name="indexmap::map::IndexMap&lt;*,*,*&gt;">
+    <DisplayString>{{ len={core.entries.len} }}</DisplayString>
+    <Expand>
+      <ExpandedItem>core.entries</ExpandedItem>
+    </Expand>
+  </Type>
+  <Type Name="indexmap::Bucket&lt;*,tuple$&lt;&gt;&gt;">
+    <DisplayString>{key}</DisplayString>
+    <Expand>
+      <Item Name="[key]">key</Item>
+      <Item Name="[hash]">hash.__0,d</Item>
+    </Expand>
+  </Type>
+  <Type Name="indexmap::Bucket&lt;*,*&gt;">
+    <DisplayString>{{ key={key}, value={value} }}</DisplayString>
+    <Expand>
+      <Item Name="[key]">key</Item>
+      <Item Name="[value]">value</Item>
+      <Item Name="[hash]">hash.__0,d</Item>
+    </Expand>
+  </Type>
+</AutoVisualizer>
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,6 +3,11 @@
 #![warn(rust_2018_idioms)]
 #![doc(html_root_url = "https://docs.rs/indexmap/1/")]
 #![no_std]
+#![cfg_attr(
+    feature = "debugger_visualizer",
+    feature(debugger_visualizer),
+    debugger_visualizer(natvis_file = "../debug_metadata/indexmap.natvis")
+)]
 
 //! [`IndexMap`] is a hash table where the iteration order of the key-value
 //! pairs is independent of the hash values of the keys.
--- a/tests/debugger_visualizer.rs
+++ b/tests/debugger_visualizer.rs
@@ -1,0 +1,44 @@
+use debugger_test::debugger_test;
+use indexmap::IndexMap;
+use indexmap::IndexSet;
+
+#[inline(never)]
+fn __break() {}
+
+#[debugger_test(
+    debugger = "cdb",
+    commands = r#"
+.nvlist
+dx set
+dx map
+"#,
+    expected_statements = r#"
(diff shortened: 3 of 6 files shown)


src/lib.rs
```diff
@@ -3,6 +3,11 @@
 #![warn(rust_2018_idioms)]
 #![doc(html_root_url = "https://docs.rs/indexmap/1/")]
 #![no_std]
+#![cfg_attr(
+    feature = "debugger_visualizer",
+    feature(debugger_visualizer),
+    debugger_visualizer(natvis_file = "../debug_metadata/indexmap.natvis")
+)]
```
cuviper, line 10: I'm personally very wary of using any unstable features in a stable crate, even behind an optional feature.

debug_metadata/indexmap.natvis
```diff
@@ -0,0 +1,29 @@
+<AutoVisualizer xmlns="http://schemas.microsoft.com/vstudio/debugger/natvis/2010">
+  <Type Name="indexmap::set::IndexSet&lt;*,*&gt;">
```
cuviper, line 2: This is more a comment toward RFC 3191, I suppose, but I'm surprised to see unmangled path matching here, especially since the `indexmap` root could refer to multiple instances in the same binary, like 1.x and 2.x. I suppose to be more precise, you would need one of the listed alternative or future possibilities that puts the attribute directly on the type declaration.

tests/debugger_visualizer.rs
```diff
@@ -0,0 +1,44 @@
...
+[0]              : 0 [Type: indexmap::Bucket<i32,tuple$<> >]
+[1]              : 1 [Type: indexmap::Bucket<i32,tuple$<> >]
+[98]             : 98 [Type: indexmap::Bucket<i32,tuple$<> >]
+
+map              : { len=0x63 } [Type: indexmap::map::IndexMap<i32,i32,std::collections::hash::map::RandomState>]
+[len]            : 0x63 [Type: unsigned __int64]
+[0]              : { key=0, value=0 } [Type: indexmap::Bucket<i32,i32>]
+[1]              : { key=1, value=-1 } [Type: indexmap::Bucket<i32,i32>]
+[2]              : { key=2, value=-2 } [Type: indexmap::Bucket<i32,i32>]
+[98]             : { key=98, value=-98 } [Type: indexmap::Bucket<i32,i32>]
```
cuviper, line 27: Is the `[N]` part showing the positional index? I expect so, but it's confusing that you used the same key values.

ridwanabdillahi: `[N]` is the positional index of each `Bucket` in the underlying `Vec` structure. I can update the examples to ease the confusion.

debug_metadata/indexmap.natvis
ridwanabdillahi: The type name for a Natvis definition matches against the debuginfo names embedded into the PDB. This does **not** contain any of the crate disambiguators you mention above in terms of version numbers. 

In the case where a binary contains multiple versions of a given crate, if both crates define the same type, the Natvis visualizer will match against both types.

Natvis does have support for `Conditions` which means certain fields or definitions can be gated upon whether or not a condition is met. This can help alleviate the issue of multiple versions of the same type.

src/lib.rs
ridwanabdillahi: I understand, the goal is to push the `debugger_visualizer` feature to stabilization relatively soon by having it provide value to crates in the wild today as with any unstable feature.

As you said it is gated behind an optional feature but you as the crate maintainer would have to potentially take on an extra cost which I fully understand could be worrisome.

tests/debugger_visualizer.rs
cuviper: Thanks, the new examples are clear, and more fun! :slightly_smiling_face:

src/lib.rs
cuviper: Well, I gave approval for CI to run, at least, so it can be a proof of concept for the feature.

ridwanabdillahi: Great! I see the CI was green and tests passed. Thanks for approving the CI job.

ridwanabdillahi: @cuviper Do you have any reservations about taking in this change now that the CI shows as green?

cuviper: Passing CI is necessary, but doesn't change the maintenance aspect. If the unstable feature changes, as it is allowed, then we'll have to update again to fix our own CI. That's especially true if any users (on nightly) start depending on it (providing "value in the wild").

For ad hoc demonstrations in the meantime, you could use cargo's patch mechanisms to point at your git fork.

cuviper:
The feature was stabilized in rust-lang/rust#108668, headed for Rust 1.71. Does this PR need updating for that? At the very least, we should not need the `feature(debugger_visualizer)` attribute anymore, but I don't know if there were any functional changes along the way.

PR #248, opened 2022-11-18
Add `iter_mut` for `IndexSet`

Seungwoo-Yu:
This patch adds iter_mut method for IndexSet to match the API for HashSet

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -157,6 +157,9 @@ impl<K, V> Bucket<K, V> {
     fn key_ref(&self) -> &K {
         &self.key
     }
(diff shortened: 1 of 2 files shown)


cuviper:
But `HashSet` does not have `iter_mut()`, unless I missed something...

Seungwoo-Yu:
> But `HashSet` does not have `iter_mut()`, unless I missed something...

Oops sorry. Never mind, it was Vector.
I needed iter_mut method to manipulate elements while ensuring that they are unique and preserves insertion order and solved problem using into_iter method and assigning it again.

PR #256, opened 2023-01-30
Improve feature flag documentation

rjsberry:
In this PR I've added a couple of commits to:

1. Use the [`doc_cfg`](https://doc.rust-lang.org/beta/unstable-book/language-features/doc-cfg.html) feature to improve readability of certain items in the crate which are locked behind feature flags.
2. Update the crate-level documentation for what the feature flags actually do.

I'm hoping this will improve clarity on the `serde` feature; it wasn't immediately clear to me that the `serde` feature enabled `Serialize` and `Deserialize` implementations, and that the functions in `serde_seq` were actually an alternative implementation for `IndexMap` specifically.

I also think that we should consider moving the `serde_seq` module from the crate root into the `map` module (similar to how the `rayon` module is organised under the `map` and `set` modules). As this module is specific to `IndexMap` I think it would be better namespaced under the `map` module. This will also help users that Ctrl+f for "serde" on the docs.rs landing page find the feature-flag documentation immediately.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ no-dev-version = true
 tag-name = "{{version}}"
 
 [package.metadata.docs.rs]
-all-features = true
+all-features = ["arbitrary", "quickcheck", "serde", "rayon"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [workspace]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ no-dev-version = true
 tag-name = "{{version}}"
 
 [package.metadata.docs.rs]
-all-features = ["arbitrary", "quickcheck", "serde", "rayon"]
+features = ["arbitrary", "quickcheck", "serde", "rayon"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [workspace]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,8 @@ no-dev-version = true
 tag-name = "{{version}}"
 
 [package.metadata.docs.rs]
-features = ["arbitrary", "quickcheck", "serde", "rayon"]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
 
 [workspace]
 members = ["test-nostd", "test-serde"]
--- a/src/map.rs
+++ b/src/map.rs
@@ -1118,7 +1118,6 @@ where
 
 #[cfg(feature = "std")]
 #[cfg_attr(docsrs, doc(cfg(feature = "std")))]
-
 impl<K, V, const N: usize> From<[(K, V); N]> for IndexMap<K, V, RandomState>
 where
     K: Hash + Eq,
--- a/src/map.rs
+++ b/src/map.rs
@@ -5,6 +5,10 @@ mod core;
 mod iter;
 mod slice;
 
+#[cfg(feature = "serde")]
+#[cfg_attr(docsrs, doc(cfg(feature = "serde")))]
+pub mod serde_seq;
+
 #[cfg(test)]
 mod tests;
 
--- a/src/map.rs
+++ b/src/map.rs
@@ -148,6 +148,7 @@ where
 }
 
 #[cfg(feature = "std")]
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -53,7 +53,8 @@ no-dev-version = true
 tag-name = "{{version}}"
 
 [package.metadata.docs.rs]
-features = ["arbitrary", "quickcheck", "serde", "rayon"]
+all-features = true
```
cuviper, line 56: I don't want all features because `rustc-rayon` is pseudo-internal.

src/rayon/map.rs
```diff
@@ -2,8 +2,8 @@
 //!
 //! You will rarely need to interact with this module directly unless you need to name one of the
 //! iterator types.
-//!
-//! Requires crate feature `"rayon"`
+
+#![cfg_attr(docsrs, doc(cfg(feature = "rayon")))]
```
cuviper, line 6: I think it would be better for both maintenance and coverage if we keep these next to the actual `#[cfg(feature)]` in question. In this case, that would be out in `lib.rs` on the entire `mod rayon`.

src/rayon/set.rs
```diff
@@ -2,8 +2,8 @@
 //!
 //! You will rarely need to interact with this module directly unless you need to name one of the
 //! iterator types.
-//!
-//! Requires crate feature `"rayon"`.
+
+#![cfg_attr(docsrs, doc(cfg(feature = "rayon")))]
```
cuviper, line 6: The one `mod rayon` attribute can cover all of these.

src/serde_seq.rs
```diff
@@ -17,8 +17,8 @@
 //!     // ...
 //! }
 //! ```
-//!
-//! Requires crate feature `"serde"`
+
+#![cfg_attr(docsrs, doc(cfg(feature = "serde")))]
```
cuviper, line 21: You already put this one on the outside in `lib.rs`.

cuviper (changes requested):
> I also think that we should consider moving the serde_seq module from the crate root into the map module (similar to how the rayon module is organised under the map and set modules). As this module is specific to IndexMap I think it would be better namespaced under the map module.

That's a good suggestion -- please do move it!

rjsberry:
Thanks for the review, I've resolved all your comments.

I've pushed the review fixes as two new commits. If you'd prefer feel free to squash to two commits (documentation update, `serde_seq` module moved) before merging.

src/map.rs
```diff
@@ -1112,6 +1117,8 @@ where
 }
 
 #[cfg(feature = "std")]
+#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
+
```
cuviper, line 1121: Stray blank line...

```suggestion
```

cuviper (changes requested):
Could you also add this to `src/arbitrary.rs`? And I suppose those should be added to the `src/lib.rs` feature list too.

src/map/serde_seq.rs
```diff
@@ -35,8 +33,6 @@ use crate::IndexMap;
 ///
 /// This behaves like [`crate::serde_seq`] for `IndexMap`, serializing a sequence
```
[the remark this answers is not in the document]
rjsberry: We can tackle that in a separate PR if you want to make the internal change? I do agree it would make sense to move the `Slice` impls. If we move to `map/serde.rs` and `set/serde.rs` though, to keep it consistent it might also be worth moving `rayon/map.rs` to `map/rayon.rs` and `rayon/set.rs` to `set/rayon.rs`.

rjsberry:
RE intra doc links, the links to `serde`, `arbitrary` and `quickcheck` traits break if the documentation is not built with the feature that enables those crates as dependencies. This isn't an issue for docs.rs but might produce some odd results if users try and build the documentation locally.

cuviper:
> RE intra doc links, the links to `serde`, `arbitrary` and `quickcheck` traits break if the documentation is not built with the feature that enables those crates as dependencies.

That is unfortunate, but I don't see a good way around it. I think docs.rs is the most important though.

PR #261, opened 2023-05-30
Implement Default for iterators

xfix:
This matches `Default` implementations for HashMap and HashSet introduced in Rust 1.70 (https://github.com/rust-lang/rust/pull/99929/).

### DIFF
--- a/src/map/iter.rs
+++ b/src/map/iter.rs
@@ -89,6 +89,12 @@ impl<K: fmt::Debug, V: fmt::Debug> fmt::Debug for Iter<'_, K, V> {
     }
 }
 
+impl<K, V> Default for Iter<'_, K, V> {
+    fn default() -> Self {
+        Self { iter: [].iter() }
+    }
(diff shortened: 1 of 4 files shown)


src/map/iter.rs
```diff
@@ -438,6 +480,14 @@ impl<K, V: fmt::Debug> fmt::Debug for ValuesMut<'_, K, V> {
     }
 }
 
+impl<K, V> Default for ValuesMut<'_, K, V> {
```
cuviper, line 483: Hmm, oddly `std` left this out, but I think it's fine.

cuviper:
> This matches Default implementations for HashMap and HashSet introduced in Rust 1.70

That actually did *not* include `hash_map`/`hash_set` iterators, but between the precedent in btree and slice iterators, I think it's still ok for us to add it here. They're trivial anyway.

cuviper (approved):

PR #262, opened 2023-06-06
Upgrade to hashbrown 0.14 (MSRV 1.64)

daxpedda:
See https://github.com/rust-lang/hashbrown/blob/v0.14.0/CHANGELOG.md.
Bumps the MSRV to 1.64.

Related #243.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -26,7 +26,7 @@
     However, exactness only applies to the direct capacity for items, while the
     raw hash table still follows its own rules for capacity and load factor.
 
-  - The `hashbrown` dependency has been updated to version 0.13.
+  - The `hashbrown` dependency has been updated to version 0.14.
 
   - The `serde_seq` module has been moved from the crate root to below the
     `map` module.
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,6 +1,6 @@
 - 2.0.0 (pending)
 
-  - **MSRV**: Rust 1.61.0 or later is now required.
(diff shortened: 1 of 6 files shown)


src/map/core/raw.rs
```diff
@@ -172,7 +172,7 @@ impl<'a, K, V> OccupiedEntry<'a, K, V> {
         // SAFETY: This is safe because it can only happen once (self is consumed)
         // and map.indices have not been modified since entry construction
         let index = unsafe { self.map.indices.remove(self.raw_bucket) };
```
cuviper, line 174: I would rather see this instead of `index.0`:
```suggestion
        let (index, _slot) = unsafe { self.map.indices.remove(self.raw_bucket) };
```

cuviper (changes requested):
Thanks, I was just looking into this too... :)

RELEASES.md
```diff
@@ -1,6 +1,6 @@
 - 2.0.0 (pending)
 
-  - **MSRV**: Rust 1.61.0 or later is now required.
+  - **MSRV**: Rust 1.64.0 or later is now required.
```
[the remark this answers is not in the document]
daxpedda: Something is wrong with GitHub today ... I've already done this, if you go to "Files changed" you will see it.

src/map/core/raw.rs
daxpedda: Done.

cuviper (approved):

PR #271, opened 2023-08-10
Upgrade to arbitrary 1.3

fornwall:
See https://github.com/gfx-rs/naga/pull/2426

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ bench = false
 [dependencies]
 equivalent = { version = "1.0", default-features = false }
 
-arbitrary = { version = "1.0", optional = true, default-features = false }
+arbitrary = { version = "1.3", optional = true, default-features = false }
 quickcheck = { version = "1.0", optional = true, default-features = false }
 serde = { version = "1.0", optional = true, default-features = false }
 rayon = { version = "1.5.3", optional = true }


cuviper:
Why wouldn't you just bump to arbitrary 1.3 in `naga`'s dependency? 

I also don't understand why this is a _new_ problem for you with indexmap 2...

daxpedda:
> Why wouldn't you just bump to arbitrary 1.3 in `naga`'s dependency?

True, I didn't consider this.

> I also don't understand why this is a _new_ problem for you with indexmap 2...

I just dug into this, the way `-Zminimal-versions` was used in `naga`s CI before was broken (I will provide a fix soon).

---

I guess `indexmap` doesn't need this PR anymore. It would probably be still good for `indexmap` to actually support building the minimal version it specifies (I'm happy to add a `-Zminimal-versions` run in the CI as well), if you don't deem it so, feel free to close this.

cuviper:
I'm on board with making sure that we have correctly specified our direct dependencies, but I'm not fond of checking `-Zminimal-versions` because of these problems with indirect dependencies. Indexmap *does* work fine with the API of `arbitrary 1.0.0`, but that full dependency chain needing `syn 1.0.6` is not really our problem.

cuviper:
Ooh, I hadn't noticed [`-Z direct-minimal-versions`](https://doc.rust-lang.org/cargo/reference/unstable.html#direct-minimal-versions) before -- that would be more palatable.

cuviper:
> I'm unsure though what `arbitrary` should do at this point though, yank all their old versions? `arbitary` v1.3 actually works correctly.

Nah, it doesn't need to be yanked. It's a bug, but not a showstopper, only affecting users that want `-Zminimal-versions` to work. Such users can update to their requirement to the fixed version 1.3.

And I understand that there's not also an `indexmap` version that you could update to solve this from afar, but anyone using `indexmap/arbitrary` should also have their own `arbitrary` dependency that can require newer.

Closing in favor of #272, thanks!

PR #272, opened 2023-08-11
Test `direct-minimal-versions`

daxpedda:
This builds the crate with [`direct-minimal-versions`](https://doc.rust-lang.org/cargo/reference/unstable.html#direct-minimal-versions) checking if the crate actually builds with the minimal dependency versions specified.

Using [`taiki-e/install-action`](https://github.com/marketplace/actions/install-development-tools) to install [`cargo-hack`](https://crates.io/crates/cargo-hack) specifically to use `--remove-dev-deps` to get the actual minimal versions users can use. If these CI dependencies are undesirable, I can make an out-of-workspace sub-crate that can have it's own lockfile without any dev-dependencies.

Cc #271.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,3 +92,21 @@ jobs:
         with:
           components: miri
       - run: cargo miri test
+
+  minimal-versions:
+    name: Check MSRV and minimal-versions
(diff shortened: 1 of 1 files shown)


cuviper:
Thanks!

PR #273, opened 2023-08-16
Improve explanation about `IndexSet`'s complexity.

your-diary:
This PR fixes [this comment in #187](https://github.com/bluss/indexmap/pull/187#issuecomment-1678785827).

In addition to just adding an explanation only about `get_index_of()` method, I added `Complexity` section to explain `IndexSet<T, S>` is internally just an `IndexMap<T, (), S>`.


cuviper:
Thanks!

PR #280, opened 2023-10-19
Use `hashbrown::HashTable` instead of `RawTable`

cuviper:


cuviper:
While this is functional, and gets rid of most `unsafe` (except our `Slice` casts), it also regresses some performance.

Issue #281, opened 2023-10-23
`binary_search` and `partition_point` on `Slice`


Eh2406:
Would it be possible to add `binary_search` and `partition_point` and friends on `Slice`?

Particular use case I have is an optimization of `retain`. I currently have `index_map.retain(|_, v| foo(&*v) < bar)` which works correctly. I happen to know that this `index_map` is sorted by `foo(&*v)`. I think it would be more efficient to do `let cut = index_map.partition_point(foo(&*v) < bar); index_map.truncate(cut)` (adjusted for off by one errors). But did not see the tools to do so.

cuviper:
I think those methods should be fine on `IndexMap` and `IndexSet` as well as their respective `Slice`s.

This could also be a building block for `insert_sorted` per #89.

Are you interested in implementing this? I think it should be straightforward to wrap the normal slice methods.

Eh2406:
I can give it a little time, happy to give it a try. Any pointers for to get started? Any gotchas to watch out for?

cuviper:
I don't think there's anything weird -- e.g. here's one for `Slice<K, V>`:

```rust
    pub fn binary_search_by<'a, F>(&'a self, mut f: F) -> Result<usize, usize>
    where
        F: FnMut(&'a K, &'a V) -> Ordering,
    {
        self.entries.binary_search_by(move |p| f(&p.key, &p.value))
    }
```

Eh2406:
I started by implement in them on `map` and `set` using `with_entries` inspired by the implementation of [`sort_by_cached_key`](https://github.com/bluss/indexmap/commit/8b7f398b2218c6da431d4cecf69ffe91ff066deb#diff-110e61eb735b79c762510f02f157a12094bcf2b9a15d2ed767532a0e3d5f7bafR747), which did not go well. Implementing them on slices went very smoothly, then the `map`/`set` can defer to `self.as_slice().foobar`.

Do you want the full documentation copied from std?

What level of testing seems appropriate?

PR #282, opened 2023-10-24
add binary_search and friends

Eh2406:
This is a start on fixing #281.

I copied the implementation of these methods from std, except for `binary_search_by` where I used std directly. Alternatively we could have all implementations directly forward to `.as_slice().foobar` or `.entries.foobar`,  if you would prefer that.

I left off the plane `binary_search{_keys}` on set and map, because they already provide a `O(1)` way to look up the index for a key. I left them in on `Slice` because they do not.

For documentation I adapted the first sentence of std, and provided a link to std.

how would you like to see these tested?

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -796,12 +796,11 @@ where
 
     /// Search over a sorted map for a key.
     ///
-    /// Returns the position where that key is present, or the position where it can be inserted to maintain the sort.
-    /// See [`slice::binary_search`] for more details.
+    /// Returns the position where that key is present, or the position where it can be inserted to
+    /// maintain the sort. See [`slice::binary_search`] for more details.
     ///
-    /// Computes in **O(log(n))** time,
-    /// which is notably less scalable than looking the key up using [`get_index_of`],
-    /// but this can also position missing keys.
+    /// Computes in **O(log(n))** time, which is notably less scalable than looking the key up
+    /// using [`get_index_of`][IndexMap::get_index_of], but this can also position missing keys.
     pub fn binary_search_keys(&self, x: &K) -> Result<usize, usize>
     where
         K: Ord,
@@ -811,8 +810,8 @@ where
 
     /// Search over a sorted map with a comparator function.
     ///
-    /// Returns the position where that value is present, or the position where it can be inserted to maintain the sort.
-    /// See [`slice::binary_search_by`] for more details.
+    /// Returns the position where that value is present, or the position where it can be inserted
+    /// to maintain the sort. See [`slice::binary_search_by`] for more details.
     ///
     /// Computes in **O(log(n))** time.
     #[inline]
@@ -825,8 +824,8 @@ where
 
     /// Search over a sorted map with an extraction function.
     ///
-    /// Returns the position where that value is present, or the position where it can be inserted to maintain the sort.
-    /// See [`slice::binary_search_by_key`] for more details.
+    /// Returns the position where that value is present, or the position where it can be inserted
+    /// to maintain the sort. See [`slice::binary_search_by_key`] for more details.
     ///
     /// Computes in **O(log(n))** time.
     #[inline]
--- a/src/map.rs
+++ b/src/map.rs
@@ -797,7 +797,7 @@ where
     /// Search over a sorted map with a comparator function.
     ///
     /// Returns the position where that value is present, or the position where it can be inserted to maintain the sort.
-    /// see [`slice::binary_search_by`] for more details.
+    /// See [`slice::binary_search_by`] for more details.
     ///
     /// Computes in **O(log(n))** time.
     #[inline]
@@ -811,7 +811,7 @@ where
     /// Search over a sorted map with an extraction function.
     ///
     /// Returns the position where that value is present, or the position where it can be inserted to maintain the sort.
-    /// see [`slice::binary_search_by_key`] for more details.
+    /// See [`slice::binary_search_by_key`] for more details.
     ///
     /// Computes in **O(log(n))** time.
     #[inline]
@@ -826,7 +826,7 @@ where
     /// Returns the index of the partition point of a sorted map according to the given predicate
     /// (the index of the first element of the second partition).
     ///
-    /// see [`slice::partition_point`] for more details.
+    /// See [`slice::partition_point`] for more details.
     ///
     /// Computes in **O(log(n))** time.
     #[must_use]
--- a/src/map.rs
+++ b/src/map.rs
@@ -794,6 +794,46 @@ where
         });
     }
 
+    /// Search over a sorted map with a comparator function.
+    ///
+    /// Returns the position where that value is present, or the position where can be inserted to maintain the sort.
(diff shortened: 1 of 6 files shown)


cuviper:
> Alternatively we could have all implementations directly forward to `.as_slice().foobar` or `.entries.foobar`, if you would prefer that.

I would, yeah, at least for `partition_point` since that's less trivial. The rest can go either way.

> I left off the plane `binary_search{_keys}` on set and map, because they already provide a `O(1)` way to look up the index for a key. I left them in on `Slice` because they do not.

That's an interesting point! I wonder if small enough maps/sets might still win by binary search, but if you're too small then even a linear search might be fastest. But anyway, if someone really wants it they can use `binary_search_by` the same way our impl would, and we could choose to add it later.

> how would you like to see these tested?

We haven't been great about that, but let's at least add sanity checks to `map/tests.rs` and `set/tests.rs`. But we don't need to be very thorough when it's relying directly on the standard library, and then type checking already proves the remaining differences.

Eh2406:
I will work on it. I have PackagingCon for then next few days a may not have much time.

Eh2406:
If I had been more awake you would have gone shorter tests. :-P I was not up to determining interesting test cases. So instead you got tests from STD copied in and adapted. They may be more thorough than you asked for.

cuviper:
Same vibe as "If I had more time, I would have written a shorter letter." :)

It's a lot, but that's fine.

src/map.rs
```diff
@@ -794,6 +794,46 @@ where
         });
     }
 
+    /// Search over a sorted map with a comparator function.
+    ///
+    /// Returns the position where that value is present, or the position where can be inserted to maintain the sort.
+    /// see [slice::binary_search_by] for more details.
+    /// **O(log(n))**
```
cuviper, line 801: Polishing like this is applicable to most of the docs.
```suggestion
    /// Returns the position where that value is present, or the position where it can be inserted to maintain the sort.
    /// See [`slice::binary_search_by`] for more details.
    ///
    /// Computes in **O(log(n))** time.
```

src/map.rs
```diff
@@ -794,6 +794,46 @@ where
...
+    #[inline]
+    pub fn binary_search_by_key<'a, B, F>(&'a self, b: &B, f: F) -> Result<usize, usize>
+    where
+        F: FnMut(&'a K, &'a V) -> B,
+        B: Ord,
+    {
+        self.as_slice().binary_search_by_key(b, f)
+    }
+
+    /// Returns the index of the partition point or a sorted map according to the given predicate
```
cuviper, line 824:
```suggestion
    /// Returns the index of the partition point of a sorted map according to the given predicate
```

src/map/slice.rs
```diff
@@ -201,6 +201,59 @@ impl<K, V> Slice<K, V> {
     pub fn into_values(self: Box<Self>) -> IntoValues<K, V> {
         IntoValues::new(self.into_entries())
     }
+
+    /// Search over a sorted map for a value.
```
cuviper, line 205:
```suggestion
    /// Search over a sorted map for a key.
```

(I'm tempted to say "key value", but "value" has a particular meaning in maps.)

src/map/tests.rs
```diff
@@ -447,3 +447,224 @@ fn iter_default() {
     assert_default::<ValuesMut<'static, K, V>>();
     assert_default::<IntoValues<K, V>>();
 }
+
+#[test]
+fn test_binary_search_by() {
+    // addaped from stds test for binary_search
```
cuviper, line 453:
```suggestion
    // adapted from std's test for binary_search
```

src/set/tests.rs
```diff
@@ -543,3 +543,142 @@ fn iter_default() {
...
+    let b: IndexSet<_> = [1, 2, 4, 5, 6, 8, 9].into();
+    assert_eq!(b.partition_point(|&x| x < 7), 5);
+    assert_eq!(b.partition_point(|&x| x < 0), 0);
+
+    let b: IndexSet<_> = [1, 3, 3, 3, 7].into();
+    assert_eq!(b.partition_point(|&x| x < 0), 0);
+    assert_eq!(b.partition_point(|&x| x < 1), 0);
+    assert_eq!(b.partition_point(|&x| x < 2), 1);
+    assert_eq!(b.partition_point(|&x| x < 3), 1);
+    assert_eq!(b.partition_point(|&x| x < 4), 2); // diff from std as set the duplicates keys
```
cuviper, line 679: This comment seems to be missing a word? Maybe you meant:
```suggestion
    assert_eq!(b.partition_point(|&x| x < 4), 2); // diff from std as set merges the duplicate keys
```

cuviper (changes requested):
Thanks - the implementation looks good, just need some cleanup in the docs and a few comments.

src/map/slice.rs
Eh2406: So does "key". Perhaps "item"? Let me see what I can do…

cuviper: In this particular case, you *are* looking at the map key.

Eh2406: Sorry, I figured that out only after posting my comment.

cuviper:
> > I left off the plane `binary_search{_keys}` on set and map, because they already provide a `O(1)` way to look up the index for a key. I left them in on `Slice` because they do not.
> 
> That's an interesting point! I wonder if small enough maps/sets might still win by binary search, but if you're too small then even a linear search might be fastest. But anyway, if someone really wants it they can use `binary_search_by` the same way our impl would, and we could choose to add it later.

Actually, there's still a gap there because binary search also finds the insertion point for missing items. So I think it's good to have the doc comment about scalability, but we should still want the map/set methods for the `Err` case.

I'll go ahead and add this myself while I'm thinking about it...

cuviper:
Thanks again! I'll put together a release soon.

cuviper:
FYI, I did publish this as 2.1.0 the other day.

I'd be interested to see where you use this, if that code is open.

Eh2406:
It is for https://github.com/pubgrub-rs/pubgrub, although the particular performance experiments where this would be useful have not yet pulled their weight. Many data structures have to "backtrack" to reset their state as of some previous point in time. For each individual data structure there are many ways this can be achieved.
- The data structure can be "cleared" and reconstructed from other data sources.
- Items in the data structure can include a "timestamp" and then the reset is `.retain(|_, v| v.timestamp < reset_to_timestamp)`
- The data structure can be maintained in an ordered fashion and reset by truncating newer modifications. If you need to scan the entire data structure to decide where to cut, then this is the same as the previous. You can have an auxiliary data structure recording `timestamp -> cutpoint`, for `O(1)` time reset but additional allocations. Using `partition point` to make a cut, gives `O(ln(size))` reset with no additional space. Which sure seems tempting.

The particular example of this pattern that inspired the PR was [`contradicted_incompatibilities`](https://github.com/bluss/indexmap/issues/281), which currently uses the first strategy but could easily use the last one. I will definitely drop you a link when/if progress toward using this PR is achieved.

PR #287, opened 2023-11-06
Implement the Merge Crate for IndexMap

### LINKED ISSUE #286: Implement (behind a feature branch) the Merge Crate
I need this [crate](https://docs.rs/merge/0.1.0/merge/) derive for my project, and thought I would quickly drop by to write the issue and try to implement it. I don't see any reason why you would not do this, as it can all be hidden behind a feature flag for the people that don't need it.

RuboGubo:
I made sure to include tests and docs, with everything passing.

### The doc of the function:
Note: If any of the rows have the same key, then they wont be merged.
the values on the right will be preserved.
```rust 
use indexmap::IndexMap;
// You need the merge crate, this crate does not re-export it.
use merge::Merge;

let expected: IndexMap<_, i32> = [("Hello", 1), ("Hi", 50), ("Ello", 2)]
    .into_iter()
    .collect();

let mut a: IndexMap<_, i32> = [("Hello", 1), ("Hi", 50)]
    .into_iter()
    .collect();

let b: IndexMap<_, i32> = [("Ello", 2), ("Hi", 2)]
    .into_iter()
    .collect();

a.merge(b);
// Note how `("Hi", 2)` in `b` wont replace `("Hi", 50)` in `a`.
// Things in `b` will be ordered after `a`
assert_eq!(expected, a);
```

The feature was also gated behind a feature flag, specifically `merge`.
All crates that I added are configured to be non-std.

Should fix #286

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -9,6 +9,9 @@ mod slice;
 #[cfg_attr(docsrs, doc(cfg(feature = "serde")))]
 pub mod serde_seq;
 
+#[cfg(feature = "merge")]
+use merge::Merge;
+
 #[cfg(test)]
 mod tests;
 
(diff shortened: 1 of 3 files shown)


src/map.rs
```diff
@@ -88,6 +91,44 @@ pub struct IndexMap<K, V, S> {
...
 }
 
+#[cfg(feature = "merge")]
+impl<K, V, S> Merge for IndexMap<K, V, S>
+where
+    K: Hash + Eq,
+    S: BuildHasher,
+{
+    /// Note: If any of the rows have the same key, then they wont be merged.
+    /// the value on the right will be preserved.
```
cuviper, line 101: This is contradicted by the example, where you show that `a.merge(b)` preserves the left values.

But either way *could* be desired, so I don't think it makes sense for us to choose. Users can derive `Merge` with their own annotation to deal with `IndexMap` in the way that works best for them.

RuboGubo: I will fix the typo in that bit of the comment, in regards to order, the merge crate does actually set a precedent, namely in the options section, it specifies that the left value should be used.

From the test section of the merge crate:
```rust
fn test<T: std::fmt::Debug + Merge + PartialEq>(expected: T, mut left: T, right: T) {
    left.merge(right);
    assert_eq!(expected, left);
}

#[derive(Debug, Merge, PartialEq)]
struct S {
    field1: Option<usize>,
}

impl S {
    pub fn new(field1: Option<usize>) -> S {
        S { field1 }
    }
}

test(S::new(Some(1)), S::new(Some(1)), S::new(Some(2)));
```

When specialization lands in the rust compiler, the user will even be able to overwrite it with their own implementation. It's just something that most people will do in the same way, so it makes sense to implement it for everyone.

cuviper: Ok, but still, `Option` is the *only* provided implementation. As I mentioned in #286, they did not choose the same for `Vec` or `HashMap`, which would be more directly applicable to `IndexMap`.

You don't need specialization -- just derive your type with a custom `#[merge(strategy = "your_func")]` on your fields.

RuboGubo: You... Make a good point :(
This is the first time I've tried to contribute to an open source project, I guess I got a bit ahead of myself. Sorry for the inconvenience.

cuviper: It's okay! It may not feel like it, but the discussion is a contribution as well, so please don't let this particular outcome deter you.

PR #290, opened 2023-12-11
Add method to get a map entry by index

bkragl:
This change adds a method `get_index_entry` to `IndexMap`, which is the by-index counterpart to `entry`. The returned struct `IndexedEntry` provides the same methods as `OccupiedEntry`, except `replace_key`.

A concrete use case where I found this useful is having a `map: IndexMap<_, Vec<_>>` with the invariant that each `Vec` is non-empty. Imagine we want to pick an entry at a random index and pop a value from the `Vec`. Without this change we'd have to use `get_mut_index` and `swap_remove_index`, and ensure that we use the same index. With `get_index_entry` we can write:

```rust
let mut e = map.get_index_entry(index).unwrap();
let v = e.get_mut();
let x = v.pop().unwrap();
if v.is_empty() {
    e.swap_remove();
}
```

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -32,7 +32,7 @@ use alloc::vec::Vec;
 #[cfg(feature = "std")]
 use std::collections::hash_map::RandomState;
 
-use self::core::IndexMapCore;
+use self::core::{IndexMapCore, IndexedEntry};
 use crate::util::{third, try_simplify_range};
 use crate::{Bucket, Entries, Equivalent, HashValue, TryReserveError};
 
@@ -421,6 +421,18 @@ where
     pub fn entry(&mut self, key: K) -> Entry<'_, K, V> {
         let hash = self.hash(&key);
         self.core.entry(hash, key)
+    }
+
+    /// Get an entry in the map by index for in-place manipulation.
+    ///
+    /// Valid indices are *0 <= index < self.len()*
+    ///
+    /// Computes in **O(1)** time.
+    pub fn get_index_entry(&mut self, index: usize) -> Option<IndexedEntry<'_, K, V>> {
+        if index >= self.len() {
+            return None;
+        }
+        Some(IndexedEntry::new(&mut self.core, index))
     }
 
     /// Return `true` if an equivalent to `key` exists in the map.
--- a/src/map.rs
+++ b/src/map.rs
@@ -423,18 +423,6 @@ where
         self.core.entry(hash, key)
     }
 
-    /// Get an entry in the map by index for in-place manipulation.
-    ///
-    /// Valid indices are *0 <= index < self.len()*
-    ///
-    /// Computes in **O(1)** time.
-    pub fn get_index_entry(&mut self, index: usize) -> Option<IndexedEntry<'_, K, V>> {
-        if index >= self.len() {
-            return None;
-        }
-        Some(IndexedEntry::new(&mut self.core, index))
-    }
-
     /// Return `true` if an equivalent to `key` exists in the map.
     ///
     /// Computes in **O(1)** time (average).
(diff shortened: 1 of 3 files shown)


src/map.rs
```diff
@@ -423,6 +423,18 @@ where
         self.core.entry(hash, key)
     }
 
+    /// Get an entry in the map by index for in-place manipulation.
+    ///
+    /// Valid indices are *0 <= index < self.len()*
+    ///
+    /// Computes in **O(1)** time.
+    pub fn entry_at(&mut self, index: usize) -> Option<EntryAt<'_, K, V>> {
```
cuviper, line 431: The name is okay, but just to bikeshed a little, what do you think about `get_index_entry` and `IndexedEntry`? I think that may fit a little better with `get_index` and `get_index_mut`, and "get" methods in general returning `Option`, as well as similarity to the other adjective-`Entry` types.

src/map/core.rs
```diff
@@ -754,9 +754,128 @@ impl<K: fmt::Debug, V> fmt::Debug for VacantEntry<'_, K, V> {
     }
 }
 
+/// A view into an occupied entry in a `IndexMap` obtained by index.
+///
+/// This struct is constructed from the `entry_at` method on `IndexMap`.
+// SAFETY: We have a mutable reference to the map, which keeps the index
+// valid and pointing to the correct entry.
```
cuviper, line 761: We don't need a "SAFETY" comment, because this is all safe code. In `OccupiedEntry`, which I assume you were copying, it's concerned with the validity of the `unsafe` access through its raw bucket.

If you still want a comment about the correctness of the index, I'd move that to the field and drop the "SAFETY" connotation.

src/map/core.rs
```diff
@@ -754,9 +754,128 @@ impl<K: fmt::Debug, V> fmt::Debug for VacantEntry<'_, K, V> {
...
+
+    /// Gets a reference to the entry's value in the map.
+    pub fn get(&self) -> &V {
+        &self.map.entries[self.index].value
+    }
+
+    /// Gets a mutable reference to the entry's value in the map.
+    ///
+    /// If you need a reference which may outlive the destruction of the
+    /// `Entry` value, see `into_mut`.
```
cuviper, line 785: The name `Entry` should be updated here.

src/map/core.rs
```diff
@@ -754,9 +754,128 @@ impl<K: fmt::Debug, V> fmt::Debug for VacantEntry<'_, K, V> {
...
+    ///
+    /// Computes in **O(n)** time (average).
+    pub fn shift_remove_entry(self) -> (K, V) {
+        self.map.shift_remove_index(self.index).unwrap()
+    }
+
+    /// Remove the key, value pair stored in the map for this entry, and return the value.
+    ///
+    /// **NOTE:** This is equivalent to `.swap_remove()`.
+    pub fn remove(self) -> V {
```
cuviper, line 832: I think we should leave out `remove` and `remove_entry`, since we're not exactly trying to mirror the hash map API in this case. (And we're talking about putting a deprecation warning on these methods elsewhere.)

src/map/core.rs
```diff
@@ -754,9 +754,128 @@ impl<K: fmt::Debug, V> fmt::Debug for VacantEntry<'_, K, V> {
...
+    ///
+    /// **NOTE:** This is equivalent to `.swap_remove_entry()`.
+    pub fn remove_entry(self) -> (K, V) {
+        self.swap_remove_entry()
+    }
+}
+
+impl<K: fmt::Debug, V: fmt::Debug> fmt::Debug for EntryAt<'_, K, V> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct(stringify!(OccupiedEntry))
```
cuviper, line 868:
```suggestion
        f.debug_struct(stringify!(EntryAt))
```

(or whatever name we land on)

The `stringify!` is weird, but the others are doing that too -- I wonder why...

Maybe the `index` is worth printing too?

cuviper (changes requested):
This does seem like a reasonable API to add, thanks!

src/map.rs
bkragl: I dig `IndexedEntry`, and also renamed the method to `get_index_entry`. Some other options for the method would be `entry_index` (fits the existing `entry` method and `get`/`get_index` pattern, but feels odd) or `indexed_entry` (fits the return type). I don't have a strong opinion, but I'm happy to do another revision if somebody else has.

src/map/core.rs
bkragl: Makes sense. If the deprecation doesn't happen, these methods could also be added later if desired.

src/map/core.rs
bkragl: Good catch and good point with the index.

Yeah, idk, I thought that `stringify!` would require some defined symbol and help in case the type is renamed, but that does not seem to be the case.

PR #292, opened 2023-12-26
Remove doc `html_root_url`

zachs18:
The `html_root_url` currently links to an old version of the docs, so if you run `cargo doc --no-deps` on a downstream package that depends on `indexmap@2.1.0`, intra-doc links to `indexmap` types will resolve to `indexmap@1.9.3` on docs.rs.


cuviper:
Oops! We should just remove this altogether, as it's no longer recommended: https://github.com/rust-lang/api-guidelines/pull/230

zachs18:
Okay, I've updated the PR to remove the attribute instead of updating it.

PR #294, opened 2024-01-13
Add splicing iterators

### LINKED ISSUE #274: Feature request: Similar method to `Vec.splice`
I'm trying to drain a range and then insert multiple items into that gap, where the number of remove items is not always equal to inserted item, basically similar to [`Vec.splice`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.splice). If that is not possible, there could also be some method that could insert an item to a particular index, so the behavior of splice could be replicated.

Currently, I do the following to replicate the behavior:
```rust
let end_part = inner_index.split_off(position);
inner_index.extend(new_items.chain(end_part.into_iter().skip(n_removed)));
```

cuviper:
`IndexMap::splice` and `IndexSet::splice` will drain the specified
range, and then insert items from an iterator to replace that range.
This is like `Vec::splice`, except if any new keys match the retained
entries outside the range, then those keep their relative position.

Resolves #274.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -14,7 +14,7 @@ mod tests;
 
 pub use self::core::{Entry, IndexedEntry, OccupiedEntry, VacantEntry};
 pub use self::iter::{
-    Drain, IntoIter, IntoKeys, IntoValues, Iter, IterMut, Keys, Values, ValuesMut,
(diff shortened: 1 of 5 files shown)


cuviper:
It might be possible to optimize this with some `unsafe` hacking, but I think this is pretty reasonable for now.

PR #296, opened 2024-01-16
Relax some over-constrained impl bounds

cuviper:
Most of these are removing `K: Hash + Eq` and/or `S: BuildHasher` where
we don't actually need to hash anything.

In some cases, we only get away with that because we have a saved hash
in each entry `Bucket`. I *suspect* we still kept the constraints as API
protection, in case we ever wanted to change our mind and stop saving
the hash. However, there are a few other unconstrained methods that also
need the saved hash, like `swap_remove_index` going back long before
version 1.0 of the crate, so that ship has sailed.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -316,13 +316,7 @@ impl<K, V, S> IndexMap<K, V, S> {
             hash_builder: self.hash_builder.clone(),
         }
     }
-}
 
-impl<K, V, S> IndexMap<K, V, S>
-where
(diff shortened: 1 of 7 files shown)


cuviper:
> However, there are a few other unconstrained methods that also
> need the saved hash, like `swap_remove_index` going back long before
> version 1.0 of the crate, so that ship has sailed.

Reference: commit 67a04b20b7ee83f1299f9531418adc79f4065a90

PR #306, opened 2024-02-01
Releases: Adjust markdown format

Turbo87:
When services like https://github.com/renovatebot/renovate open [dependency update pull requests](https://github.com/rust-lang/crates.io/pull/8043) they try to find a corresponding changelog entry to let the developers know what changes this dependency update includes.

The current format of the `RELEASES.md` file with its nested lists appears to be uncommon enough to not get recognized by at least renovatebot.

This changes the format to use `h2` elements for the versions, which should allow renovatebot to pick up the contents and generally renders in a slightly nicer way :D

Best reviewed in "ignore whitespace" mode: https://github.com/indexmap-rs/indexmap/pull/306/files?w=1 😉 

/cc @cuviper

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,159 +1,161 @@
-- 2.2.2
+# Releases
 
-  - Added indexing methods to raw entries: `RawEntryBuilder::from_hash_full`,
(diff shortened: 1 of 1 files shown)


cuviper:
Sounds good, thanks!

PR #308, opened 2024-02-03
Add map and set `extract_if`

cuviper:


src/map/core/raw.rs
```diff
@@ -143,3 +157,80 @@ impl<'a, K, V> RawTableEntry<'a, K, V> {
...
+                let dest = base.add(new_len);
+                src.copy_to(dest, tail_len);
+                new_len += tail_len;
+            }
+            self.map.entries.set_len(new_len);
+        }
+
+        if new_len != old_len {
+            // We don't keep track of *which* items were extracted, so reindex everything.
+            self.map.rebuild_hash_table();
```
cuviper, line 195: While it would be nice to reuse `vec::ExtractIf`, I haven't found a way to do this fixup for the indices afterward, because that `vec::ExtractIf` would be holding `&mut` on the entries, excluding even a raw pointer elsewhere.

PR #323, opened 2024-03-20
Opt-in mutable access on IndexSet, release 2.2.6

### LINKED ISSUE #322: MutableKeys for IndexSet
Would it be possible to add a way to mutate keys in Set like you can Map?

cuviper:
This adds `set::MutableValues` as an opt-in trait for accessing `IndexSet` values by mutable reference, akin to `map::MutableKeys`. The implementation simply forwards to the map implementation while discarding the `()` inner "value".

Unlike `IndexMap`, `IndexSet` doesn't have regular `get_full_mut` or `get_index_mut` methods, so the "2" suffix here isn't really necessary, but I thought it might be better to stay consistent with `MutableKeys` on that.

Closes #322

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "indexmap"
 edition = "2021"
-version = "2.2.5"
+version = "2.2.6"
 documentation = "https://docs.rs/indexmap/"
 repository = "https://github.com/indexmap-rs/indexmap"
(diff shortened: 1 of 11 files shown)


cuviper:
@Morganamilo would this work for you?

Morganamilo:
This is what I was looking for thanks.

Issue #67, opened 2018-02-26
impl Hash for IndexSet, IndexMap


alexheretic:
I think it would be useful for `IndexSet`, like `Vec`, to implement `Hash`. Technically there's no reason `IndexMap` couldn't either, though I'm not sure how useful it would be practically.

It should be a pretty simple addition, are there downsides?

bluss:
A hash and equality combination that is consistent needs to be present. I doubt we can or want to do that, because we also want eq rules to be the same as for hashmap(?)

alexheretic:
Yep it's one or the other. I hadn't thought about wanting equality to disregard order. I'd have expected otherwise but perhaps it's simpler this way.

Thanks for the clarification.

cuviper:
The open PR rust-lang/rust#48366 is proposing `Hash` for the standard `HashMap` and `HashSet`, using a commutative operation to combine item hashes.  Currently that's just `wrapping_add`, but it sounds like they want to explore stronger options.

cuviper:
I've been thinking about reintroducing `ordermap` as a newtype wrapper that **does** consider the order for `PartialEq` and `Eq`, also adding `Hash`, `PartialOrd`, and `Ord`, and now I've published that:
https://crates.io/crates/ordermap/0.5.0

PR #329, opened 2024-06-09
Add hash implementation

fabianboesiger:
There were several requests to implement `Hash` for `IndexMap` and `IndexSet`. The problem with implementing `Hash` is that it should be consistent with the `Eq` implementation. The `Eq` implementation is based on containing equivalent items *regardless of order*, requiring the hash to stay the same regardless of order. To fix this issue, one comment https://github.com/indexmap-rs/indexmap/issues/67#issuecomment-368576339 suggests using a commutative operation to combine the hashes of the individual entries.

This draft is an implementation of this proposal.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1415,3 +1415,22 @@ where
     S: BuildHasher,
 {
 }
+
+#[cfg(feature = "std")]
+impl<K, V, S> Hash for IndexMap<K, V, S>
+where
+    K: Hash + std::fmt::Debug,
+    V: Hash + std::fmt::Debug,
+    S: BuildHasher,
+{
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let hash = self.as_slice().iter().map(|(k, v)| {
+            let mut hasher = std::hash::DefaultHasher::new();
+            k.hash(&mut hasher);
+            v.hash(&mut hasher);
+            hasher.finish()
+        }).fold(0, u64::wrapping_add);
+
+        state.write_u64(hash);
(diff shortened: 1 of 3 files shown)


cuviper:
That comment also linked to https://github.com/rust-lang/rust/pull/48366, which was closed due to that hash approach being rather weak.

What is your use-case? Would you be able to work with a newtyped key that _does_ consider the order of items?

fabianboesiger:
Yes, I'm aware that the solution is not perfect, but as many people seem to request it, it might still make sense to consider it. 

I'm also aware that the `.as_slice()` API exists where `Slice` implements `Hash`, however, this is still very inconvenient when we want to derive `Hash` for a struct that contains an `IndexMap`.

It is also possible to add an implementation for newtyped keys that do consider the order of items, I can look into that if sou decide to add this feature.

cuviper:
> but as many people seem to request it

Not so many that we _need_ to do something, IMO, especially since the standard `HashMap` did not.

> [...] when we want to derive `Hash` for a struct that contains an `IndexMap`.

I am still skeptical that it makes sense to do that. We can figure out ways to create a `Hash`, ordered or not, but that doesn't necessarily make it useful. Do you have a concrete use-case that you can share?

fabianboesiger:
My use case is I'd like to create sets of sets. I could use BTreeSets for that, but this is a bit too slow.

I think I will close this PR for now and create my own wrapper struct that I can use for my use case. Feel free to reopen this in case it is needed in the future.

PR #333, opened 2024-07-26
Use `serde_derive` via `derive` feature on `serde`

waywardmonkeys:
This is the recommended way to do this in the `serde` docs.


waywardmonkeys:
This is for the dev dependency only, so won't impact dependent crates of `indexmap`, so it won't run afoul of https://github.com/serde-rs/serde/issues/2584

cuviper:
As you note by that issue, that recommendation is being reconsidered. I don't see any point in changing here -- what would we be *fixing* by doing so?

waywardmonkeys:
I think best to close this!

PR #335, opened 2024-07-29
Alow Mutating Entry keys and iterator keys

### LINKED ISSUE #334: Expanding mut key API
`toml_edit` presents `Table` as a mapping between `Key` and `Value` with `KeyMut` for providing access to the subset of key functionality that the user can modify.  Internally, this is stored as `IndexMap<String, (Key, Value)>`, where `Key` contains a `String`, among other data.

`Key.get().clone()` to get the `String` key is showing up in my benchmarks in parsing a real-world `Cargo.toml` file.  I'm exploring my options for avoiding that

 I hadn't noticed `MutableKeys` before and it seems like a good start by allowing me to instead do `IndexMap<Key, Value>`.  Some parts I'm missing for maintaining my `KeyMut` access for users:
- `MutableKeys::iter_full_mut2`
- `MutableOccupiedEntrykeys::get_full_mut2`

There may be more as I've not fully gone through my compilation errors.  I'm more wondering if this direction is viable before I continue down this path.

I'd be willing to contribute these if there is interest.

epage:
See toml-rs/toml#765 for an example use case

Fixes #334

### DIFF
--- a/src/map/mutable.rs
+++ b/src/map/mutable.rs
@@ -1,7 +1,8 @@
 use core::hash::{BuildHasher, Hash};
 
 use super::{
-    Bucket, Entries, Entry, Equivalent, IndexMap, IndexedEntry, OccupiedEntry, VacantEntry,
+    Bucket, Entries, Entry, Equivalent, IndexMap, IndexedEntry, IterMut2, OccupiedEntry,
+    VacantEntry,
 };
 
 /// Opt-in mutable access to [`IndexMap`] keys.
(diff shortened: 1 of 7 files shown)


cuviper:
> See [toml-rs/toml#765](https://github.com/toml-rs/toml/pull/765) for an example use case

Any performance numbers to share from that change?

epage:
Machine has been pretty jittery though,. so using callgrind I get
- Before: 8,879,894
- After: 8,678,207

This also cleaned up the code a lot.

src/map/mutable.rs
```diff
@@ -80,8 +90,77 @@ where
...
+    fn key_mut(&mut self) -> &mut Self::Key;
+}
+
+/// Opt-in mutable access to [`Entry`] keys.
+///
+/// See [`MutableEntryKey`] for more information.
+impl<K, V> MutableEntryKey for Entry<'_, K, V> {
+    type Key = K;
+
+    /// Gets a nutable reference to the entry's key, either within the map if occupied,
```
cuviper, line 118:
```suggestion
    /// Gets a mutable reference to the entry's key, either within the map if occupied,
```

cuviper:
Thanks! I've published 2.3.0 with these changes.

PR #364, opened 2024-12-30
remove fxhash example and dep

hkBst:
remove fxhash example and dep as [fxhash is unmaintained](https://github.com/rustsec/advisory-db/issues/2185)


cuviper:
Maybe it's worth replacing that with [`rustc-hash`](https://crates.io/crates/rustc-hash)? That one is maintained for the Rust compiler itself, and still uses the "Fx" prefix, although it's not exactly the same algorithm anymore.

cuviper:
Ok, we can leave it with just `fnv`.

PR #367, opened 2025-01-13
Improve panic messages and add `#[track_caller]` attribute to functions that may panic

savannstm:
Currently, `indexmap` panic messages aren't really helpful, and this PR adds `#[track_caller]` attributes to functions that may panic, for better understanding where to actually look for the panic, and uses the same panic messages that standard collections do with more details.
Seems like `#[track_caller]` is compatible with no_std, and it shouldn't break anything.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -299,6 +299,7 @@ impl<K, V, S> IndexMap<K, V, S> {
     ///
     /// ***Panics*** if the starting point is greater than the end point or if
     /// the end point is greater than the length of the map.
+    #[track_caller]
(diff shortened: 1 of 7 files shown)


src/map.rs
```diff
@@ -1325,7 +1346,7 @@ where
     ///
     /// ***Panics*** if `key` is not present in the map.
     fn index(&self, key: &Q) -> &V {
```
cuviper, line 1348: Just clarifying for future reference -- our `Index` and `IndexMut` implementations don't need `#[track_caller]` because the trait definition already applies it.

cuviper (approved):
Seems like a good idea, thanks!

PR #371, opened 2025-01-21
ci: use cargo-nextest with miri

cuviper:


cuviper:
From 5m 2s to 4m 23s -- I was hoping for more reduction, since my own system runs `cargo miri nextest run` under a minute, but I guess the standard GitHub runners just aren't as powerful. Oh well, it's still an improvement!

PR #372, opened 2025-01-27
Drop the internal `rustc-rayon` impls

cuviper:
While this looks like a breaking change, it was always documented in `Cargo.toml` as internal-only for use in the compiler:

> Internal feature, only used when building as part of rustc,
> not part of the stable interface of this crate.

That's flipped around in [rustc-rayon#14] and [rust#136149].

[rustc-rayon#14]: https://github.com/rust-lang/rustc-rayon/pull/14
[rust#136149]: https://github.com/rust-lang/rust/pull/136149


cuviper:
> That's flipped around in [rustc-rayon#14](https://github.com/rust-lang/rustc-rayon/pull/14) and [rust#136149](https://github.com/rust-lang/rust/pull/136149).

I'll wait for the latter to merge to make sure that works out.

PR #377, opened 2025-02-21
Upgrade dev-dependency `rand` to 0.9

waywardmonkeys:


cuviper (approved):
Wow, I'm surprised we didn't hit *any* of the breaking changes. Thanks for the update!

waywardmonkeys:
> Wow, I'm surprised we didn't hit _any_ of the breaking changes. Thanks for the update!

Me too.

PR #380, opened 2025-02-26
feat: add no_std macros `indexmap_with_default` and `indexset_with_default`

### LINKED ISSUE #379: We need `indexmap` and `indexset`-like macros that don't require `std`
As seen here https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/crates/proof-of-sql/src/base/map.rs#L8-L23
we can have `indexmap` and `indexset`-like macros without having to have `std`.

iajoiner:
Closes #379.
I'm adding `indexmap_with_default` and `indexset_with_default` which should be no-std.

Thanks @JayWhite2357 for providing a solution in [Proof of SQL](https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/crates/proof-of-sql/src/base/map.rs) which we are upstreaming!

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,3 +1,40 @@
+/// Create an [`IndexMap`][crate::IndexMap] from a list of key-value pairs
+/// and a `BuildHasherDefault`-wrapped custom hasher.
+///
+/// ## Example
+///
+/// ```
+/// use indexmap::indexmap_with_default;
+/// use fnv::FnvHasher;
+///
+/// let map = indexmap_with_default!{
+///     FnvHasher;
+///     "a" => 1,
+///     "b" => 2,
+/// };
+/// assert_eq!(map["a"], 1);
+/// assert_eq!(map["b"], 2);
+/// assert_eq!(map.get("c"), None);
+///
+/// // "a" is the first key
+/// assert_eq!(map.keys().next(), Some(&"a"));
+/// ```
+#[macro_export]
+macro_rules! indexmap_with_default {
+    ($H:ty; $($key:expr => $value:expr,)+) => { $crate::indexmap_with_default!($H; $($key => $value),+) };
+    ($H:ty; $($key:expr => $value:expr),*) => {{
+        let builder = ::core::hash::BuildHasherDefault::<$H>::default();
+        const CAP: usize = <[()]>::len(&[$({ stringify!($key); }),*]);
+        #[allow(unused_mut)]
+        // Specify your custom `H` (must implement Default + Hasher) as the hasher:
+        let mut map = $crate::IndexMap::with_capacity_and_hasher(CAP, builder);
+        $(
+            map.insert($key, $value);
+        )*
+        map
+    }};
+}
+
 #[cfg(feature = "std")]
 #[cfg_attr(docsrs, doc(cfg(feature = "std")))]
 #[macro_export]
@@ -33,6 +70,43 @@ macro_rules! indexmap {
             map
         }
     };
+}
+
+/// Create an [`IndexSet`][crate::IndexSet] from a list of values
+/// and a `BuildHasherDefault`-wrapped custom hasher.
+///
+/// ## Example
+///
+/// ```
+/// use indexmap::indexset_with_default;
+/// use fnv::FnvHasher;
+///
+/// let set = indexset_with_default!{
+///     FnvHasher;
+///     "a",
+///     "b",
+/// };
+/// assert!(set.contains("a"));
+/// assert!(set.contains("b"));
+/// assert!(!set.contains("c"));
+///
+/// // "a" is the first value
+/// assert_eq!(set.iter().next(), Some(&"a"));
+/// ```
+#[macro_export]
+macro_rules! indexset_with_default {
+    ($H:ty; $($value:expr,)+) => { $crate::indexset_with_default!($H; $($value),+) };
+    ($H:ty; $($value:expr),*) => {{
+        let builder = ::core::hash::BuildHasherDefault::<$H>::default();
+        const CAP: usize = <[()]>::len(&[$({ stringify!($value); }),*]);
+        #[allow(unused_mut)]
+        // Specify your custom `H` (must implement Default + Hash) as the hasher:
+        let mut set = $crate::IndexSet::with_capacity_and_hasher(CAP, builder);
+        $(
+            set.insert($value);
+        )*
+        set
+    }};
 }
 
 #[cfg(feature = "std")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -98,7 +98,8 @@
 //!   [`with_capacity_and_hasher`][IndexMap::with_capacity_and_hasher] instead.
 //!   A no-std compatible hasher will be needed as well, for example
 //!   from the crate `twox-hash`.
-//! - Macros [`indexmap!`] and [`indexset!`] are unavailable without `std`.
+//! - Macros [`indexmap!`] and [`indexset!`] are unavailable without `std`. Use
+//!   the macros [`indexmap_with_default!`] and [`indexset_with_default!`] instead.
 
 #![cfg_attr(docsrs, feature(doc_cfg))]
 


src/macros.rs
```diff
@@ -63,7 +60,8 @@ macro_rules! indexset {
             // Note: `stringify!($value)` is just here to consume the repetition,
             // but we throw away that string literal during constant evaluation.
             const CAP: usize = <[()]>::len(&[$({ stringify!($value); }),*]);
-            let mut set = $crate::IndexSet::with_capacity(CAP);
+            #[allow(unused_mut)]
+            let mut set = $crate::IndexSet::with_capacity_with_hasher(CAP, <_>::default());
```
cuviper, line 64:
```suggestion
            let mut set = $crate::IndexSet::with_capacity_and_hasher(CAP, <_>::default());
```

iajoiner:
@cuviper Really thanks for letting you know! How do you think we can modify it to fix the issue then?

iajoiner:
Will something like this be acceptable?
```rust
#[macro_export]
macro_rules! indexmap {
    ($S:ty; $($key:expr => $value:expr),*) => {{
        const CAP: usize = <[()]>::len(&[$({ stringify!($key); }),*]);
        #[allow(unused_mut)]
        // Specify your custom S (must implement Default) as the hasher:
        let mut map = $crate::IndexMap::<_, _, $S>::with_capacity_and_hasher(CAP, <$S>::default());

        $(
            map.insert($key, $value);
        )*

        map
    }};
    // fallback without hasher type specified
    ($($key:expr => $value:expr),*) => {{
        use std::collections::hash_map::RandomState;
        const CAP: usize = <[()]>::len(&[$({ stringify!($key); }),*]);
        #[allow(unused_mut)]
        let mut map = $crate::IndexMap::<_, _, RandomState>::with_capacity_and_hasher(
            CAP, 
            RandomState::default()
        );

        $(
            map.insert($key, $value);
        )*

        map
    }};
}
```
This way if `std` is enabled we can allow the old macro without any change. Otherwise we require the hasher to be specified.

cuviper:
I don't know -- this is why the macros are currently hard-coded. :wink:

`FromIterator` does have open `S`, and now that this trait is in the prelude, it will work reasonably well with array iterators:

```rust
// this still needs *something* to drive the type inference for `S`
let map = IndexMap::from_iter([(1, 'a'), (2, 'b'), (3, 'c')]);
```

cuviper:
> Will something like this be acceptable?
> 
> ```rust
> #[macro_export]
> macro_rules! indexmap {
>     ($S:ty; $($key:expr => $value:expr),*) => {{
>         [...]
>     }};
>     // fallback without hasher type specified
>     ($($key:expr => $value:expr),*) => {{
>         [...]
>     }};
> }
> ```

If this works without any change to the existing tests, then yes, I think we can do that!

JayWhite2357:
Another solution could be this:
```rust
#[cfg(feature = "std")]
let mut map = $crate::IndexMap::with_capacity(CAP);
#[cfg(not(feature = "std"))]
let mut map = $crate::IndexMap::with_capacity_and_hasher(CAP, <_>::default());
```

My syntax might be off, but something like that should work. It would require specifying the type explicitly when `std` is off, but it would be essentially identical semantics otherwise. Very similar to the `from_iter`.

cuviper:
No, we can't just alternate it because `indexmap` could be a dependency of multiple parts of the build tree, perhaps with and without the "std" feature enabled, so it will be unified to enabled. Then the crate which wanted it with some other hasher will be broken.

JayWhite2357:
> No, we can't just alternate it because `indexmap` could be a dependency of multiple parts of the build tree, perhaps with and without the "std" feature enabled, so it will be unified to enabled. Then the crate which wanted it with some other hasher will be broken.

Ah, yes.

iajoiner:
@cuviper @JayWhite2357 Alternatively maybe we can separate `indexmap_with_hasher` / `indexset_with_hasher` from `indexmap` and `indexset`. Then users who have picked a particular hasher can just define their own `indexmap` and `indexset` using `indexmap_with_hasher` / `indexset_with_hasher`.

iajoiner:
Now use cases need to look like this if we use a different hasher
```rust
use ahash::AHasher;
use indexmap::indexmap_with_hasher;

fn main() {
    let map = indexmap_with_hasher! {
        AHasher; "a" => 1, "b" => 2,
    };
    assert_eq!(map["a"], 1);
    assert_eq!(map["b"], 2);
    assert_eq!(map.get("c"), None);
}
```

iajoiner:
I have tried this with both `ahash` and `twox-hash` and `DefaultHasher` is covered in doc tests so I think we are good to go.

src/macros.rs
```diff
@@ -1,3 +1,39 @@
...
+/// assert_eq!(map.get("c"), None);
+///
+/// // "a" is the first key
+/// assert_eq!(map.keys().next(), Some(&"a"));
+/// ```
+#[macro_export]
+macro_rules! indexmap_with_hasher {
+    ($S:ty; $($key:expr => $value:expr,)+) => { $crate::indexmap_with_hasher!($S; $($key => $value),+) };
+    ($S:ty; $($key:expr => $value:expr),*) => {{
+        type BuildS = core::hash::BuildHasherDefault<$S>;
```
cuviper, line 25: I think this is mixing terminology, although it's not helped by `HashMap`'s original terminology either.

* `S` is the type that implements `BuildHasher`. (I don't know why the letter `S` was chosen.) The `$S:ty` you're taking here is the kind of `Hasher` that an `S` would build, so maybe it should be `$H`.
* The normal `with_hasher` method takes an instance of that _builder_ `S`, not an actual hasher instance. That's already perhaps confusing, but it will muddy the waters even more if these macro `*_with_hasher` are working with the underlying `Hasher` type instead.
  * Maybe `indexmap_with_default!(AHash; key => value, ...)` is a better name? Then there's a clue that this is using `BuildHasherDefault` inside.

Bonus nit: this should use `::core::hash::...` to make sure the path is using the crate namespace, and not some local module (like our own `map/core.rs`).

src/macros.rs
```diff
@@ -1,3 +1,39 @@
...
+        const CAP: usize = <[()]>::len(&[$({ stringify!($key); }),*]);
+        #[allow(unused_mut)]
+        // Specify your custom S (must implement Default + Hasher) as the hasher:
+        let mut map = $crate::IndexMap::<_, _, BuildS>::with_capacity_and_hasher(CAP, <BuildS>::default());
```
cuviper, line 29: I don't think we need the map turbofish or type alias at all. This `<BuildS>::default()` is just giving you a `BuildS` (which is really `S`), the same as its `new()` method. So I think this can be more like:
```suggestion
        let builder = ::core::hash::BuildHasherDefault::<$H>::new();
        let mut map = $crate::IndexMap::with_capacity_and_hasher(CAP, builder);
```

cuviper (changes requested):

src/macros.rs
```diff
@@ -1,3 +1,39 @@
+/// Create an [`IndexMap`][crate::IndexMap] from a list of key-value pairs and a custom hasher
+///
+/// ## Example
+///
+/// ```
+/// use indexmap::indexmap_with_default;
+/// use std::hash::DefaultHasher;
+///
+/// let map = indexmap_with_default!{
+///     DefaultHasher;
```
cuviper, line 10: We have a dev-dep on `fnv`, so let's use that for the examples.

src/macros.rs
```diff
@@ -35,6 +71,42 @@ macro_rules! indexmap {
     };
 }
 
+/// Create an [`IndexSet`][crate::IndexSet] from a list of key-value pairs and a custom hasher
```
cuviper, line 74:
```suggestion
/// Create an [`IndexSet`][crate::IndexSet] from a list of values
/// and a `BuildHasherDefault`-wrapped custom hasher.
```

src/macros.rs
```diff
@@ -1,3 +1,40 @@
+/// Create an [`IndexMap`][crate::IndexMap] from a list of key-value pairs  
+/// and a `BuildHasherDefault`-wrapped custom hasher.
```
cuviper, line 2: Trailing whitespace crept in here:
```suggestion
/// Create an [`IndexMap`][crate::IndexMap] from a list of key-value pairs
/// and a `BuildHasherDefault`-wrapped custom hasher.
```

I'm surprised `rustfmt` doesn't try to change that, but I guess it's something to do with markdown line-wrapping. We don't want to _force_ rendering in this particular way.

src/macros.rs
```diff
@@ -1,3 +1,40 @@
...
+/// assert_eq!(map.get("c"), None);
+///
+/// // "a" is the first key
+/// assert_eq!(map.keys().next(), Some(&"a"));
+/// ```
+#[macro_export]
+macro_rules! indexmap_with_default {
+    ($H:ty; $($key:expr => $value:expr,)+) => { $crate::indexmap_with_default!($H; $($key => $value),+) };
+    ($H:ty; $($key:expr => $value:expr),*) => {{
+        let builder = ::core::hash::BuildHasherDefault::<$H>::new();
```
cuviper, line 26: Oh, I didn't realize `new()` was *just* stabilized in Rust 1.85. (https://github.com/rust-lang/rust/issues/123197)

Better go back to `default()` then.
```suggestion
        let builder = ::core::hash::BuildHasherDefault::<$H>::default();
```

src/macros.rs
```diff
@@ -35,6 +72,43 @@ macro_rules! indexmap {
...
+/// assert!(!set.contains("c"));
+///
+/// // "a" is the first value
+/// assert_eq!(set.iter().next(), Some(&"a"));
+/// ```
+#[macro_export]
+macro_rules! indexset_with_default {
+    ($H:ty; $($value:expr,)+) => { $crate::indexset_with_default!($H; $($value),+) };
+    ($H:ty; $($value:expr),*) => {{
+        let builder = ::core::hash::BuildHasherDefault::<$H>::new();
```
cuviper, line 100:
```suggestion
        let builder = ::core::hash::BuildHasherDefault::<$H>::default();
```

cuviper (changes requested):

iajoiner:
@cuviper Really thanks! It's ready again haha.

cuviper (approved):
Looks good, thanks!

PR #385, opened 2025-04-03
docs: provide link to `core::hash::BuildHasherDefault` in `macros.rs` docs

iajoiner:
Currently the doc for `indexmap_with_default` and `indexset_with_default` mention `core::hash::BuildHasherDefault` but don't provide any hyperlink to [its page](https://doc.rust-lang.org/beta/core/hash/struct.BuildHasherDefault.html). We can easily fix this so that it is easier for people to use the two macros.


cuviper (approved):
LGTM, thanks!

PR #389, opened 2025-04-30
Add `#[track_caller]` to all functions that are marked with `***Panics***` and can potentially panic

rakivo:


src/map.rs
```diff
@@ -1388,6 +1388,7 @@ where
     /// Returns a reference to the value corresponding to the supplied `key`.
     ///
     /// ***Panics*** if `key` is not present in the map.
+    #[track_caller]
```
cuviper, line 1391: Last time, we decided this wasn't needed because the trait already has it:
https://github.com/indexmap-rs/indexmap/pull/367#discussion_r1921673139

src/map.rs
```diff
@@ -1473,15 +1475,16 @@ impl<K, V, S> Index<usize> for IndexMap<K, V, S> {
...
-            })
-            .1
+        if let Some((_, value)) = self.get_index(index) {
+            value
+        } else {
+            panic!(
+                "index out of bounds: the len is {len} but the index is {index}",
+                len = self.len()
+            );
+        }
```
cuviper, line 1487: I think you're right that this closure was in the way though, so this is a good change.

src/map/core.rs
```diff
@@ -80,6 +80,7 @@ fn update_index(table: &mut Indices, hash: HashValue, old: usize, new: usize) {
 /// and without regard for duplication.
 ///
 /// ***Panics*** if there is not sufficient capacity already.
+#[track_caller]
 fn insert_bulk_no_grow<K, V>(indices: &mut Indices, entries: &[Bucket<K, V>]) {
```
cuviper, line 84: This is an internal function, and the pre-condition is entirely in our own control. If this ever fails, I think it would be confusing to the user to have their caller location reported, when they really should report a bug to us. So, I don't think we should add this one.

src/map.rs
rakivo: Oh, I didn't know that the trait already has it, thank you for the clarification

cuviper (changes requested):
Please do remove those on `Index` and `IndexMut` traits, and on the internal function.

rakivo:
Ok, here you go

src/map/core.rs
```diff
@@ -183,7 +183,6 @@ impl<K, V> IndexMapCore<K, V> {
         }
     }
 
-    #[track_caller]
```
cuviper, line 186: Only your addition on the internal `insert_bulk_no_grow` should have been removed. All of these other existing internal instances are needed to track public API panics. In other words, your overall PR should not show any removals, only the agreed additions.

Once you've done that, please squash your fixup commits as well. The closure change can be kept separately if you like, or it would be fine to squash that down too.

cuviper (changes requested):

src/map/core.rs
cuviper: Note: while I only commented on this one, **everything** that existed before your PR should be restored.

PR #390, opened 2025-05-13
Fix Vulnerability RUSTSEC-2024-0402

muhrifqii:
Vulnerability found on hashbrown 0.15.0 https://osv.dev/vulnerability/RUSTSEC-2024-0402

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,12 +25,12 @@ rayon = { version = "1.9", optional = true }
 borsh = { version = "1.2", optional = true, default-features = false }
 
 [dependencies.hashbrown]
-version = "0.15.0"
+version = "0.15.3"
 default-features = false
 
 [dev-dependencies]
 itertools = "0.14"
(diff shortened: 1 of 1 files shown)


cuviper:
We never use `hashbrown/borsh` at all, so that vulnerability is irrelevant to us. If you have anything else using that feature, you should set your requirement there.

In other words, I'm closing this as "not affected".

muhrifqii:
https://github.com/indexmap-rs/indexmap/blob/1818d4140d86aeef18c515f1b060a3fa68da2708/src/lib.rs#L238

cuviper:
Even with that version, we are not enabling the vulnerable feature, so it doesn't matter. 

Also, Cargo will resolve that to any semver-compatible version -- it treats "0.15.0" the same as ">= 0.15.0, < 0.16.0". So if you have something else that requires a newer 0.15.*, indexmap will use that too.

PR #392, opened 2025-05-24
Add some missing tests

AbeZbm:
Hi,

Thanks for your time to review this PR.

By examining the existing code, we found that some tests can be added to improve the repo's overall test coverage. We've added some tests to cover missing unit tests in [slice.rs](https://github.com/indexmap-rs/indexmap/blob/main/src/map/slice.rs), [map.rs](https://github.com/indexmap-rs/indexmap/blob/main/src/map.rs) and [set.rs](https://github.com/indexmap-rs/indexmap/blob/main/src/set.rs). The tests we submitted have been carefully curated by us to ensure their behavior and effectiveness.

Thanks again for reviewing.

### DIFF
--- a/src/map/tests.rs
+++ b/src/map/tests.rs
@@ -594,6 +594,214 @@ fn iter_default() {
 }
 
 #[test]
+fn get_index_mut2() {
+    let mut map: IndexMap<i32, i32> = IndexMap::new();
+    map.insert(1, 2);
+    map.insert(3, 4);
+    map.insert(5, 6);
+
+    {
+        let (key, value) = map.get_index_mut2(0).unwrap();
+        assert_eq!(*key, 1);
+        assert_eq!(*value, 2);
+
+        *value = 7;
+    }
+    assert_eq!(map.get_index(0).unwrap().1, &7);
+
+    {
+        let (key, _) = map.get_index_mut2(0).unwrap();
+        *key = 8;
+    }
+    assert_eq!(map.get_index(0).unwrap().0, &8);
+}
+
+#[test]
+fn shift_shift_remove_index() {
+    let mut map: IndexMap<i32, i32> = IndexMap::new();
+    map.insert(1, 2);
+    map.insert(3, 4);
+    map.insert(5, 6);
+    map.insert(7, 8);
+
+    let result = map.shift_remove_index(2);
+    assert_eq!(result, Some((5, 6)));
+    assert_eq!(map.len(), 3);
+
+    let result = map.shift_remove_index(1);
+    assert_eq!(result, Some((3, 4)));
+    assert_eq!(map.len(), 2);
+
+    let result = map.shift_remove_index(2);
+    assert_eq!(result, None);
+    assert_eq!(map.len(), 2);
+}
+
+#[test]
+fn shift_remove_entry() {
+    let mut map: IndexMap<i32, i32> = IndexMap::new();
+    map.insert(1, 2);
+    map.insert(3, 4);
+    map.insert(5, 6);
+    map.insert(7, 8);
+
+    let result = map.shift_remove_entry(&5);
+    assert_eq!(result, Some((5, 6)));
+    assert_eq!(map.len(), 3);
+
+    let result = map.shift_remove_entry(&3);
+    assert_eq!(result, Some((3, 4)));
+    assert_eq!(map.len(), 2);
+
+    let result = map.shift_remove_entry(&9);
+    assert_eq!(result, None);
+    assert_eq!(map.len(), 2);
+}
+
+#[test]
+fn shift_remove_full() {
+    let mut map: IndexMap<i32, i32> = IndexMap::new();
+    map.insert(1, 2);
+    map.insert(3, 4);
+    map.insert(5, 6);
+    map.insert(7, 8);
+
+    let result = map.shift_remove_full(&5);
+    assert_eq!(result, Some((2, 5, 6)));
+    assert_eq!(map.len(), 3);
+
+    let result = map.shift_remove_full(&3);
+    assert_eq!(result, Some((1, 3, 4)));
+    assert_eq!(map.len(), 2);
+
+    let result = map.shift_remove_full(&9);
+    assert_eq!(result, None);
+    assert_eq!(map.len(), 2);
+}
+
+#[test]
+fn sorted_unstable_by() {
+    let mut map: IndexMap<i32, i32> = IndexMap::new();
+    map.extend(vec![(1, 10), (2, 20), (3, 30), (4, 40), (5, 50)]);
+    let sorted = map.sorted_unstable_by(|a, b, c, d| d.cmp(&b));
+
+    let expected: Vec<(i32, i32)> = vec![(5, 50), (4, 40), (3, 30), (2, 20), (1, 10)];
+    let result: Vec<(i32, i32)> = sorted.map(|(k, v)| (k, v)).collect();
+    assert_eq!(result, expected);
+}
+
+#[test]
+fn into_boxed_slice() {
+    let mut map: IndexMap<i32, i32> = IndexMap::new();
+    for i in 0..100 {
+        map.insert(i, i * 10);
+    }
+    let boxed_slice: Box<Slice<i32, i32>> = map.into_boxed_slice();
+    assert_eq!(boxed_slice.len(), 100);
+    for i in 0..100 {
+        assert_eq!(boxed_slice[i], i as i32 * 10);
+    }
+}
+
+#[test]
+fn last_mut() {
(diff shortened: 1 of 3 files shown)


src/map/tests.rs
```diff
@@ -593,6 +593,214 @@ fn iter_default() {
...
+
+    let result = index_map.get_range(4..2);
+    assert!(result.is_none());
+
+    let result = index_map.get_range(2..4);
+    let slice: &Slice<i32, i32> = result.unwrap();
+    assert_eq!(slice.len(), 2);
+    for i in 0..slice.len() {
+        assert_eq!(slice.entries[i].key, i as i32 + 3);
+        assert_eq!(slice.entries[i].value, (i as i32 + 3) * 10);
```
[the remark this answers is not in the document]
AbeZbm: Thanks for your correction and thanks for your time to review my PR!😊

src/map/tests.rs
```diff
@@ -593,6 +593,219 @@ fn iter_default() {
...
+    map.insert(3, 4);
+    map.insert(5, 6);
+    map.insert(7, 8);
+    map.insert(9, 10);
+
+    let result = map.shift_remove_index(1);
+    assert_eq!(result, Some((3, 4)));
+    assert_eq!(map.len(), 4);
+
+    map.swap_remove(&5);
```
cuviper, line 632: This swap is a bad choice, because it makes the other shifts irrelevant. That is, I can change all of those other shifts to swap (e.g. as if the shift was implemented incorrectly), and the expected values here still pass!

Just forget this line, and the rest could be more reasonable.

- `shift_remove_index(1) -> (3, 4)`
- `shift_remove_index(1) -> (5, 6)` -- because it **shifted** to fill the gap)
- `shift_remove_index(2) -> (9, 10)` -- because the last was never swapped, only shifted down twice, and it also tests that shift-removing the last entry is fine
- `shift_remove_index(2) -> None`

src/map/tests.rs
```diff
@@ -593,6 +593,219 @@ fn iter_default() {
...
+    map.insert(5, 6);
+    map.insert(7, 8);
+    map.insert(9, 10);
+
+    let result = map.shift_remove_entry(&3);
+    assert_eq!(result, Some((3, 4)));
+    assert_eq!(map.len(), 4);
+
+    map.swap_remove(&5);
+    assert_eq!(map.as_slice(), &[(1, 2), (9, 10), (7, 8)]);
```
cuviper, line 657: Similarly here, the `swap_remove` makes it so we can't tell whether the first one really shifted or swapped. Remove that and you can compare the full shifted order instead.

src/map/tests.rs
```diff
@@ -593,6 +593,219 @@ fn iter_default() {
...
+    map.insert(3, 4);
+    map.insert(5, 6);
+    map.insert(7, 8);
+    map.insert(9, 10);
+
+    let result = map.shift_remove_full(&3);
+    assert_eq!(result, Some((1, 3, 4)));
+    assert_eq!(map.len(), 4);
+
+    map.swap_remove(&5);
```
cuviper, line 681: Again, don't use swap when testing shifts.

src/map/tests.rs
```diff
@@ -593,6 +593,219 @@ fn iter_default() {
...
+    let result = index_map.get_range_mut(4..2);
+    assert!(result.is_none());
+
+    let result = index_map.get_range_mut(2..4);
+    let slice: &mut Slice<i32, i32> = result.unwrap();
+    assert_eq!(slice.len(), 2);
+    assert_eq!(slice, &mut [(3, 30), (4, 40)]);
+
+    for i in 0..slice.len() {
+        slice.entries[i].value += 1;
```
cuviper, line 794:
```suggestion
        slice[i] += 1;
```

src/set/tests.rs
```diff
@@ -583,6 +583,264 @@ fn iter_default() {
...
+    assert_eq!(index_set.as_slice(), &[10, 20, 30]);
+}
+
+#[test]
+fn try_reserve_exact() {
+    let mut index_set: IndexSet<i32> = IndexSet::new();
+    index_set.insert(10);
+    index_set.insert(20);
+    index_set.insert(30);
+    assert_eq!(index_set.capacity(), 3);
```
cuviper, line 636: I think this is too dependent on the growth strategy and load factor, which are implementation details of both hashbrown and this crate combined, but it would be ok if the test calls `shrink_to_fit()` first. We still can't be sure indices table won't have higher capacity, but the entries vec will shrink to the exact capacity, and we report the lesser.

The rest of the test is ok for the same reason -- the vec will grow exactly even if the indices table doesn't.

src/set/tests.rs
```diff
@@ -583,6 +583,264 @@ fn iter_default() {
...
+    set.insert(20);
+    set.insert(30);
+    set.insert(40);
+    set.insert(50);
+
+    let result = set.shift_remove_full(&20);
+    assert_eq!(result, Some((1, 20)));
+    assert_eq!(set.len(), 4);
+
+    set.swap_remove(&30);
```
cuviper, line 656: Like I mentioned in the map tests, don't mix swaps in any of the shift tests.

src/set/tests.rs
```diff
@@ -583,6 +583,264 @@ fn iter_default() {
...
+
+#[test]
+fn drain() {
+    let mut set: IndexSet<i32> = IndexSet::new();
+    set.insert(1);
+    set.insert(2);
+    set.insert(3);
+
+    let drain = set.drain(0..2);
+    assert_eq!(drain.as_slice(), &[1, 2]);
```
cuviper, line 717: It would be worth testing the output of the drain iterator, and the state of the set afterward.

src/set/tests.rs
```diff
@@ -583,6 +583,264 @@ fn iter_default() {
...
+
+    let result = set.shift_remove_index(2);
+    assert_eq!(result, None);
+    assert_eq!(set.len(), 2);
+}
+
+#[test]
+fn sort_unstable_by() {
+    let mut set: IndexSet<i32> = IndexSet::new();
+    set.extend(vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
```
cuviper, line 694: This doesn't need a vec - just convert a direct array.

```suggestion
    let mut set: IndexSet<i32> = IndexSet::from([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
```

This or similar can be applied in a few of the new tests.

src/set/tests.rs
```diff
@@ -583,6 +583,264 @@ fn iter_default() {
...
+
+    let drain = set.drain(0..2);
+    assert_eq!(drain.as_slice(), &[1, 2]);
+}
+
+#[test]
+fn split_off() {
+    let mut set: IndexSet<i32> = IndexSet::new();
+    set.extend(vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
+    set.truncate(5);
```
cuviper, line 724:
```suggestion
    let mut set: IndexSet<i32> = IndexSet::from([1, 2, 3, 4, 5]);
```

src/set/tests.rs
```diff
@@ -583,6 +583,264 @@ fn iter_default() {
...
+    let split_set: IndexSet<i32> = set.split_off(3);
+    assert_eq!(split_set.len(), 2);
+    assert_eq!(set.len(), 3);
```
cuviper, line 727: How about checking the contents of each as well.

src/set/tests.rs
```diff
@@ -583,6 +583,264 @@ fn iter_default() {
...
+fn last() {
+    let mut set: IndexSet<i32> = IndexSet::new();
+    set.insert(1);
+    set.insert(2);
+    set.insert(3);
+    set.insert(4);
+    set.insert(5);
+    set.insert(6);
+
+    set.pop();
```
cuviper, line 803: Seems irrelevant to `insert` and immediately `pop` here.

cuviper (changes requested):

AbeZbm:
Hi cuviper,

I owe you a sincere apology for the multiple pings earlier. When I saw your activity but no review progress for this pr, I mistakenly assumed you might have missed my updates — especially since I deleted and reposted comments thinking it would refresh the notification for you. I now realize this created four disruptive pings, and I fully understand it was both unnecessary and disrespectful of your time. I realize it must have made you feel pushed or even overwhelmed. That was never my intention, and I’m truly sorry for the annoyance. 

Your latest feedback has been fully addressed in the new commit. I’d appreciate your review whenever it’s convenient for you. Thank you for your patience and for teaching me so much.

> I will review again when I get a chance, but please stop pinging so much. You've repeated nearly the same comment now 4 times, which I guess you deleted and reposted. I see no urgency just for test coverage.

PR #400, opened 2025-06-29
Implement sval2 support

Techcable:
Supports serialization ~~without recursion, allowing arbitrary nesting~~.

This library is used as part of the `log` and `value-bag` crates.


src/sval.rs
```diff
@@ -0,0 +1,36 @@
+#![cfg_attr(docsrs, doc(cfg(feature = "serde")))]
```
cuviper, line 1:
```suggestion
#![cfg_attr(docsrs, doc(cfg(feature = "sval")))]
```

Cargo.toml
```diff
@@ -51,11 +52,11 @@ sign-tag = true
 tag-name = "{{version}}"
 
 [package.metadata.docs.rs]
-features = ["arbitrary", "quickcheck", "serde", "borsh", "rayon"]
+features = ["arbitrary", "quickcheck", "serde", "borsh", "rayon", "sval"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [workspace]
-members = ["test-nostd", "test-serde"]
+members = ["test-nostd", "test-serde", "test-sval"]
```
cuviper, line 59: This test needs to be added to `.github/workflows/ci.yml` as well.

PR #402, opened 2025-07-10
Add `insert_sorted_by{,_key}` methods for map/set

### LINKED ISSUE #401: Add `insert_sorted_by` and `insert_sorted_by_key`
This would be equivalent to `binary_search_by` or `binary_search_by_key` and then `insert_before`.

4e554c4c:
Resolves #401.

### DIFF
--- a/src/map/tests.rs
+++ b/src/map/tests.rs
@@ -1226,3 +1226,27 @@ fn disjoint_indices_mut_fail_duplicate() {
         Err(crate::GetDisjointMutError::OverlappingIndices)
     );
 }
+
+#[test]
+fn insert_sorted_by_key() {
+    let values = [(-1, 8), (3, 18), (-27, 2), (-2, 5)];
+    let mut map: IndexMap<i32, i32> = IndexMap::new();
+    for (key, value) in values.into_iter() {
+        map.insert_sorted_by_key(|k, _| k.abs(), key, value);
+    }
+    let values: Vec<_> = map.iter().map(|(k,v)|(*k,*v)).collect();
+    map.sort_by_cached_key(|key, _| key.abs());
+
+    assert_eq!(values, map.into_iter().collect::<Vec<_>>());
+}
+
+#[test]
+fn insert_sorted_by() {
+    let mut values = [(1, 1), (2, 2), (3, 3), (4, 4), (5, 5)];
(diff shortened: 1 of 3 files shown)


src/set.rs
```diff
@@ -422,6 +422,56 @@ where
...
+    /// Insert the value into the set at its ordered position among values
+    /// sorted by `cmp`.
+    ///
+    /// This is equivalent to finding the position with
+    /// [`binary_search_by`][Self::binary_search_by], then calling
+    /// [`insert_before`][Self::insert_before].
+    ///
+    /// If the sorted item is found in the set, it returns the index of that
+    /// existing item and `false`. Otherwise, it inserts the new item and
+    /// returns its sorted index and `true`.
```
cuviper, line 434: This paragraph needs adjustment because existing items also move -- or maybe just leave it out like you did with the map method.

src/set.rs
```diff
@@ -422,6 +422,56 @@ where
...
+    /// Insert the value into the set at its ordered position among values
+    /// sorted by `cmp`.
+    ///
+    /// This is equivalent to finding the position with
+    /// [`binary_search_by_key`][Self::binary_search_by] with `cmp(key)`, then
+    /// calling [`insert_before`][Self::insert_before].
+    ///
+    /// If the sorted item is found in the set, it returns the index of that
+    /// existing item and `false`. Otherwise, it inserts the new item and
+    /// returns its sorted index and `true`.
```
cuviper, line 459: Same here, either update or remove this paragraph.

src/map/tests.rs
```diff
@@ -1226,3 +1226,27 @@ fn disjoint_indices_mut_fail_duplicate() {
...
+
+#[test]
+fn insert_sorted_by_key() {
+    let values = [(-1, 8), (3, 18), (-27, 2), (-2, 5)];
+    let mut map: IndexMap<i32, i32> = IndexMap::new();
+    for (key, value) in values.into_iter() {
+        map.insert_sorted_by_key(|k, _| k.abs(), key, value);
+    }
+    let values: Vec<_> = map.iter().map(|(k,v)|(*k,*v)).collect();
+    map.sort_by_cached_key(|key, _| key.abs());
```
cuviper, line 1238: I think it would be better for this to sort the original `values` as a more independent check.

PR #411, opened 2025-09-06
feat: add IndexMap::get_key_value_mut

ya7010:
Added a method whose implementation had been forgotten.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -828,7 +828,7 @@ where
         self.get_index_of(key).is_some()
     }
 
-    /// Return a reference to the value stored for `key`, if it is present,
+    /// Return a reference to the stored value for `key`, if it is present,
     /// else `None`.
     ///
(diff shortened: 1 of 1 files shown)


cuviper:
Nit: the title and commit message left out the `_mut` suffix.

This is a simple thing to add, but I notice that only `hashbrown` has this map method, not the standard library's `HashMap` nor `BTreeMap`. So it's not really missing or forgotten in terms of `std` parity, but we should probably add this everywhere. Would you be willing to propose this as an [ACP](https://std-dev-guide.rust-lang.org/development/feature-lifecycle.html)?

In the meantime, `IndexMap::get_full_mut` provides the same access, along with the index.

ya7010:
Thank you for your assistance!

Since this is my first time submitting to ACP, I'll try to find time to work on it.

PR #414, opened 2025-09-14
Depend on `serde_core` instead of `serde`

DaniPopes:
This crate does not make use of serde derive macros, thus it can depend on `serde_core` instead of `serde` to speed up users' compile times.

See the documentation of [`serde_core`](https://docs.rs/serde_core) for more details.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ equivalent = { version = "1.0", default-features = false }
 
 arbitrary = { version = "1.0", optional = true, default-features = false }
 quickcheck = { version = "1.0", optional = true, default-features = false }
-serde = { version = "1.0", optional = true, default-features = false }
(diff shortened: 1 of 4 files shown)


epage:
Beat me to this, thanks!

Cargo.toml
```diff
@@ -34,11 +34,13 @@ itertools = "0.14"
 fastrand = { version = "2", default-features = false }
 quickcheck = { version = "1.0", default-features = false }
 fnv = "1.0"
+serde = { version = "1.0", default-features = false }
```
cuviper, line 37: Why is this needed?

cuviper (commented):
Thanks!

There are also a couple doc-links in `src/lib.rs` that need updating.

cuviper (approved):

PR #427, opened 2025-12-01
Implement `AbsDiffEq`, `RelativeEq` and `UlpsEq` from `approxim` crate for `IndexMap`

stanpapa:
This PR adds `AbsDiffEq`, `RelativeEq` and `UlpsEq` from the [approxim](https://github.com/jonaspleyer/approxim) crate for `IndexMap`. The trait implementation are locked behind the `approxim` feature flag.

These traits build on top of `PartialEq` and are used to compare values with variable precision, which simplifies float comparisons.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,7 +41,7 @@
 //!   dependency that arose between [`borsh`] and `indexmap`, `borsh v1.5.6`
 //!   added an `indexmap` feature that should be used instead of enabling the
 //!   feature here.
-//! * `approxim`: Add implementations for [`AbsDiffEq`], [`RelativeEq`] and [`UlpsEq`]
+//! * `approxim`: Adds implementations for [`AbsDiffEq`], [`RelativeEq`] and [`UlpsEq`]
 //!   to [`IndexMap`].
 //!
 //! _Note: only the `std` feature is enabled by default._


stanpapa:
Thanks for taking the time and providing feedback so quickly!

> This seems like a quite niche thing to add, in part because approxim has much less usage than the original approx crate, yet we don't have that implementation either.

I agree it is quite niche. The reason for choosing approxim instead of approx is because approx does not seem to be maintained anymore and approxim has more functionality than approx.

> These traits also don't seem to be container-oriented -- they're not implemented for the standard BTreeMap or HashMap, nor even Vec. They are implemented for slices, but I'd guess that's more simd-oriented. What is the use case where you need an IndexMap itself to implement these traits?

My use case is that I would like to compare collections of molecular geometries (which are just lists of 3D coordinates). These collections are part of a bigger struct and to reduce boilerplate, we would like to derive these traits. Having IndexMap implement these traits makes that much cleaner.

> Without any new dependencies, I could imagine having a generic comparison method, something like:
>
> ```rust
> fn eq_by(&self, other: &IndexMap<K, V2, S2>, eq: impl FnMut(&V, &V2)) -> bool;
> ```
> Would that meet your needs?

That would already simplify things somewhat. Were you thinking of something likes this:
```rust
    fn eq_by<V2, S2>(&self, other: &IndexMap<K, V2, S2>, eq: impl Fn(&V, &V2) -> bool) -> bool
    where
        S2: BuildHasher,
    {
        self.len() == other.len()
            && self
                .iter()
                .all(|(key, value)| other.get(key).map_or(false, |v| eq(value, v)))
    }
```

Given that my request is admittedly too niche for this crate, I will close this PR and try to have the implementation added to approxim itself, which might make more sense in this case.

Would you like me to open a new PR for the `eq_by` function or would you rather do it yourself?

